PR #7, opened 2020-12-19
Add case keyword to rc lexer to match vis

moesasji:
The vis editor added the case keyword to the rc lexer in [this commit](https://github.com/martanne/vis/commit/97bfecaf908bbe2ae5ac87475c85cfba2fe5267f#diff-36b4b40fc374f8fb19f69b9f70dc2b6d5e0898393e6da1830d6ec23e56c2046c).

Adding this would enable to sync the lexers.


orbitalquark:
Fixed via https://github.com/orbitalquark/scintillua/commit/69d5544feefde7974cf15e52b595558c7581f98f

PR #8, opened 2020-12-20
Add Python3 async and await keywords to match vis

moesasji:
This adds the python3 async and await keywords to the Python lexer


orbitalquark:
Fixed via https://github.com/orbitalquark/scintillua/commit/50d9630e299b752e9439ec93f7ce56a191673d08

PR #9, opened 2020-12-20
Add async/await keywords to javascript

moesasji:
The vis editor added a commit adding async/await keywords to the javascript lexer in commit
36b573ad416f18e14f85d69cd5acf431b33fd48a 

This merges that change as a step towards bringing the lexers back in sync


orbitalquark:
Fixed via https://github.com/orbitalquark/scintillua/commit/3dd0b6e9abc10ce96fc5987ff931514e902862fe

PR #10, opened 2020-12-20
Add fix for crash in rest lexer from vis

moesasji:
Vis commited a fix for a crash in the rest lexer in
352404dc442896029b01a0da6f9b29e39ec5c53c to address
https://github.com/martanne/vis/issues/665

It would be good if this could be merged to enable syncing


orbitalquark:
Fixed via https://github.com/orbitalquark/scintillua/commit/f8ba01e062263bc4709c98a14d6990a198543c22

PR #12, opened 2020-12-20
Make function local in sml lexer

moesasji:
This adds part of commit 4af2de4fb0483d388c7a6e3d43d8a5b8bf78042e
in vis that corrects a function having global scope in the sml lexer


orbitalquark:
Fixed via https://github.com/orbitalquark/scintillua/commit/d7c687add9288daf0e7060f986e7a923cadd194d

PR #13, opened 2020-12-20
Add legacy lexers present in vis not upstreamed

moesasji:
This pull request merges all lexers that are present in the vis editor lexer code, but were never upstreamed. This includes some relatively popular languages. 

These are currently all in the legacy style, but my guess is that it is better to first merge and then update.


moesasji:
I now realize that this one needs more work as I forgot to include the file-associations in lpeg.properties
.

orbitalquark:
Thanks, an updated lpeg.properties would be great. Then I can work on migrating these legacy lexers and committing them.

moesasji:
I've now updated the pull request to include the file-extensions from [this file in vis](https://github.com/moesasji/vis/blob/master/lua/plugins/filetype.lua)

Note that I'm not a 100% sure that I have some of the ones that don't use a file-extension correct, but this should at least be a start.

orbitalquark:
Okay, I've converted most of these to the new lexer format and committed. The others I prefer not to include for one reason or another. Thanks for submitting all of these!

PR #14, opened 2020-12-20
Add fix for slow yaml lexer as in vis source

moesasji:
This adds a fix for "incorrect and slow yaml lexer" present in vis source as commit 79859b57fc3556c2ed74950a5a5668c64face4e3 that didn't get upstreamed.

See: [this commit](https://github.com/martanne/vis/commit/79859b57fc3556c2ed74950a5a5668c64face4e3#diff-1c0754c7810f4d45fcd1c6c8663f0edb69765675f147b6255c0a054c67de5a0d)


orbitalquark:
Fixed via https://github.com/orbitalquark/scintillua/commit/8829edbab66109ad59d5e4c6a99a149a8fd7d5de

I did not incorporate the patch directly because it does not handle keys like "foo bar" and "foo  :".

PR #15, opened 2020-12-20
Add xml lexer performance improvement used in vis

moesasji:
This adds performance improvements for the xml lexer added to vis in commits: 7e9e0a2ca868aaa214fb38a79fe71da34d6e00da and
baa51e934ce057af5b5be829d6a73a3e8b4c03d0

### DIFF
--- a/lexers/xml.lua
+++ b/lexers/xml.lua
@@ -47,13 +47,7 @@ lex:add_rule('attribute', token('attribute', identifier) * namespace^-1 *
 lex:add_style('attribute', lexer.styles.type)
 
 -- TODO: performance is terrible on large files.
-local in_tag = P(function(input, index)
-  local before = input:sub(1, index - 1)
-  local s, e = before:find('<[^>]-$'), before:find('>[^<]-$')
-  if s and e then return s > e and index or nil end
-  if s then return index end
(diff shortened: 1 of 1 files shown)


orbitalquark:
This change does not correctly handle the following input:

    <size
      width=100
      height=100>

The "width=100" line will not be highlighted correctly, but the line below it will. This is why I had to use a function for look-behind.

Note that currently there is no performance penalty because that look-behind function is not called. ('=' is not highlighted at all, and "=[number]" sequences are still highlighted, but not necessarily in the context of attributes.)

mcepl:
> This change does not correctly handle the following input:
>
> ```
> <size
> width=100
> height=100>
> ```

Just that this is not correct XML, attributes must be quoted. For my test text ([18 MB text of Bible](https://gitlab.com/crosswire-bible-society/CzeCSP/-/blob/master/bible.xml?ref_type=heads)) it seems to work just fine:

xml-element

mcepl:
All three related PRs are rebased on the current `master` in the `devel` branch, i.e. in https://git.sr.ht/~mcepl/vis/commit/92e7198cfc81

orbitalquark:

orbitalquark:
Closing per comments in #118.

PR #19, opened 2020-12-21
add copyright statement to template

moesasji:
The current template lacks a copyright statement. My guess is that it is better to add a default statement that puts it under your name and the default license. Those that want to can then add their own line.

The attached file just add the default statement at the top.


orbitalquark:
Fixed via https://github.com/orbitalquark/scintillua/commit/9dd533ed88c0da6c4c63a764c1f149ca3da7ca08

PR #20, opened 2020-12-21
Add whitespace detection to the text lexer

moesasji:
Vis has a commit 807ba6f6850bf7dbf86fd65299248f2d9cba7f75 to match whitespaces as explained [here](https://github.com/martanne/vis/commit/807ba6f6850bf7dbf86fd65299248f2d9cba7f75) 

Although I'm not a 100% sure why you would want to do this it is probably worth adding, which is what this pull request does.


orbitalquark:
Fixed via https://github.com/orbitalquark/scintillua/commit/2aae3e3f532b059c08f30a8e5f5a7d21d132e345

PR #37, opened 2021-01-30
add character matching to lisp

ErmineII:
before `#\"` would treat the rest of the file as a string, which is ugly, but now characters are highlighted


ErmineII:
(matches character literals like [these](https://www.tutorialspoint.com/lisp/lisp_characters.htm)), a more sophisticated approach matches literals like `#\newline`)

```lisp
(a literal double quote is #\" when displayed correctly)
```

```lua
(a literal double quote is #\" when displayed incorrectly)
```

orbitalquark:
Thanks! Fixed via https://github.com/orbitalquark/scintillua/commit/cf91d8c75c95c90158207cb4bc65fafce4d9a532

Issue #27, opened 2020-12-25
Heredoc formatting in the bash lexer is not correct


moesasji:
Currently vis has a set of patches contributed by @silasdb aiming to improve the formatting of heredocs: [fix ending delimiter](https://github.com/martanne/vis/commit/0eb675b33fdc309b5124b4164b5333ef4a64a1d8#diff-77caf695dadebc3270093253f1827e75126e21ad828717f313fba832fa194242) and a set to recognize <<- [here](https://github.com/martanne/vis/commit/f4f0f5b22a2e52bd0391ee7f6cd4b99c98028fe9#diff-77caf695dadebc3270093253f1827e75126e21ad828717f313fba832fa194242) and [here](https://github.com/martanne/vis/commit/e57851806592f5ea4ee48303b97a5f59731370c5#diff-77caf695dadebc3270093253f1827e75126e21ad828717f313fba832fa194242).

However even with these patches the formatting is not yet correct for all correct usages of heredoc as it doesn't deal correctly with spaces between << and the delimiter. Moreover the formatting fails when the first delimiter is being quoted to prevent variable expansion. As a result fixing this appears not that trivial.

I've attached a test-file that highlights the above problems as this is easier to show than explain in part based on [this tutorial](https://www.baeldung.com/linux/heredoc-herestring) that are worth fixing and probably a useful step to get this working correctly.

File including valid usage of heredoc: [test-bash.zip](https://github.com/orbitalquark/scintillua/files/5741771/test-bash.zip)

silasdb:
This is something I'd like to tackle.  Before trying your test, I tried to run the Scintillua test suite and I got a bunch of errors.  It seems that lexers cannot find the global lpeg (e.g.: `Failed! lexers/rhtml.lua:6: attempt to index a nil value (global 'lpeg')`).  Tests work fine if I explictly `local lpeg = require('lpeg')` for each lexer.

Could it be my environment?  I'm using NetBSD 9.1 Lua 5.3 + LPEG.  Not sure if you have this problem in GNU/Linux or another *nix.

moesasji:
I get the same error message on Slackware Current if I use lua 5.3. 
The tests complete without error if I use lua 5.1 (still the default one)

orbitalquark:
*tests.lua* should probably define `_G.lpeg`. My computer is set up to have lua5.1 execute the tests and I haven't bothered to run it with any other lua version.

silasdb:
[...]

PR #43, opened 2021-02-27
Fix bash heredocs

silasdb:
With these fixes, it can highlight the most common heredocs usage. Shell expansion rules allow some strange syntax but shell script writers usually prefer to keep it simple, so this should suffice.

This fixes issue #27.

The string pattern we use became a bit more complex than the previous version but it also covers more cases.


lexers/bash.lua
```diff
@@ -27,11 +27,20 @@ local sq_str = lexer.range("'", false, false)
...
 local ex_str = lexer.range('`')
 local heredoc = '<<' * P(function(input, index)
-  local _, e, _, delimiter = input:find(
-    '^%-?(["\']?)([%a_][%w_]*)%1[\n\r\f;]+', index)
-  if not delimiter then return end
-  _, e = input:find('[\n\r\f]+' .. delimiter, e)
-  return e and e + 1 or #input + 1
+  local s, e, minus, _, delimiter =
+    input:find('(-?)%s*(["\']?)([%w_]+)%2[^\n]*[\n\r\f;]+', index)
+  if s == index and delimiter then
```
orbitalquark: Is there a reason you removed the '^' pattern anchor and are instead checking if `s == index`? If you put the anchor back and remove the `s == index` check, does the fix still behave as expected? I ask because '^' in Lua patterns makes matching much faster in general.

lexers/bash.lua
```diff
@@ -27,11 +27,20 @@ local sq_str = lexer.range("'", false, false)
 local dq_str = lexer.range('"')
 local ex_str = lexer.range('`')
 local heredoc = '<<' * P(function(input, index)
-  local _, e, _, delimiter = input:find(
-    '^%-?(["\']?)([%a_][%w_]*)%1[\n\r\f;]+', index)
-  if not delimiter then return end
-  _, e = input:find('[\n\r\f]+' .. delimiter, e)
-  return e and e + 1 or #input + 1
+  local s, e, minus, _, delimiter =
+    input:find('(-?)%s*(["\']?)([%w_]+)%2[^\n]*[\n\r\f;]+', index)
```
orbitalquark: `[%w_]` will also match numbers. Are numbers valid delimiters? I am asking because I don't know all of the intricacies of bash.

What is the reason for `[^\n]*` before the newline/semicolon check? Are there valid characters between a beginning delimiter and the end-of-line? Again, I just don't know all the bash details.

orbitalquark (changes requested):

silasdb:
On Mon, Mar 01, 2021 at 08:18:24AM -0800, orbitalquark wrote:
>+  local s, e, minus, _, delimiter =
>+    input:find('(-?)%s*(["\']?)([%w_]+)%2[^\n]*[\n\r\f;]+', index)
>+  if s == index and delimiter then
>
>Is there a reason you removed the '^' pattern anchor and are instead checking if `s == index`? If you put the anchor back and remove the `s == index` check, does the fix still behave as expected? I ask because '^' in Lua patterns makes matching much faster in general.

Hmmmm, the old approach is indeed better and simplier.  I don't recall exactly why I changed this, but it is probably on my experiments when tweaking the vis bash lexer.

I'm going to make some more other tests before I revert this and update this pull request.

>+  local s, e, minus, _, delimiter =
>+    input:find('(-?)%s*(["\']?)([%w_]+)%2[^\n]*[\n\r\f;]+', index)
>
>`[%w_]` will also match numbers. Are numbers valid delimiters? I am asking because I don't know all of the intricacies of bash.

Yes, numbers are part of valid delimiters, like this one:

cat <<123
Hello world
123

>What is the reason for `[^\n]*` before the newline/semicolon check? Are there valid characters between a beginning delimiter and the end-of-line? Again, I just don't know all the bash details.

Bash (the Bourne shell and descendants, actually) doesn't consider trailing spaces or semicollons after the opening delimiter for compose the its name.  Because of that, we need to consume everything up to the newline character, if not, the delimiter variable ends up having the trailing spaces.

See this example (please, replace one _ by one space).

cat <<EOF___
Hello world
EOF

bash accepts this, but in order to make it highlight correctly, I need to consume everything up to the newline.

I just realized that, if I do this, '[\n\r\f;]+' probably becomes unnecessary.

BTW, there is room for simplification.  I'll simplify, make more tests, and submit this PR again.

May I add a detailed commentary just before the pattern matching?

Thanks for the review!

orbitalquark:
Thanks for the clarification. Feel free to add documentation.

mcepl:
@silasdb Is this PR still relevant? Could you finish it please?

silasdb:
On Mon, Jul 18, 2022 at 12:32:05PM -0700, Matěj Cepl wrote:
> @silasdb Is this PR still relevant? Could you finish it please?

Hi! I believe it is. Sorry for taking so long to address that. I'll finish it before August 12th, ok? Thanks!

silasdb:
On Tue, Jul 26, 2022 at 10:28:35AM -0300, Silas wrote:
> Hi! I believe it is. Sorry for taking so long to address that. I'll finish it before August 12th, ok? Thanks!

I fixed what was requested by @orbitalquark. Indeed it is now simplier and it
seems to work fine, although performing further tests might be a good idea.

I though about replacing "[\n\r\f]" for "\n" only but didn't test weird cases
where a "\r" and "\f" are used so I preferred not to change that for now.

orbitalquark:
Thanks! I'll take a look and merge when I have some time.

orbitalquark:
Implemented these changes in https://github.com/orbitalquark/scintillua/commit/956e4ce3039cdea06226933c603f579e37d2bc5e. I tested using the sample zip in #27. Thanks!

PR #44, opened 2021-02-27
Add Gleam lexer

tynanbe:
See:

- https://github.com/gleam-lang/
- https://gleam.run/

This lexer represents a first pass.


orbitalquark:
Thanks! Would you please also add the appropriate entry for this lexer to *lpeg.properties*?

orbitalquark:
Thanks again. Since your lexer is in the legacy format, I'll need some time to convert it[1] to the new format before committing. If you want to do it yourself, you can, but it's not required.

[1]: https://orbitalquark.github.io/scintillua/api.html#migrating-legacy-lexers

orbitalquark:
Fixed via https://github.com/orbitalquark/scintillua/commit/29d83f8ec0a1f4f8a8b0fd361a02935119301c59

PR #50, opened 2021-08-21
Makefile lexer updated to support multiple targets

nthirtyone:
Patch adds support for multiple targets in a single rule as POSIX standard describes them. It does not introduce support for grouped targets as they are defined by GNU Make.

I'm propagating the same (pending) change from vis editor. It was checked against vis - I used it for at least half a year so far. However, it was not tested directly with scintillua, if you find it necessary, I can look into setting something up.


orbitalquark:
Great, thanks! This has been on my TODO list for a while. Fixed via https://github.com/orbitalquark/scintillua/commit/4db1d2f83c96ccbdd97e60a5897e007a31047e16

Issue #53, opened 2021-09-12
LexLPeg.dll crash


ittegrat:
I'm new to Scintillua and I was trying the 5.1 release with SciTE 5.1.1 64bit (single file 64-bit executable). I just added the line `import lexers/lpeg` to an empty `SciTEGlobal.properties` and it crashes immediately. I also tried the 32bit version (with SciTE 32), but it doesn't seem to work (doesn't crash, but no styles rendered).
Here are my findings till now:
- Scintillua 5.1, SciTE 5.1.1: 64bit crash, 32bit doesn't work (both SciTE single executabele and SciTE full)
- Scintillua 5.0, SciTE 5.0.3: 64bit crash, 32bit doesn't work
- Scintillua 4.4.5-2, SciTE 4.4.5: 64bit crash, **32bit does work**

Can I do some tests to help solve the problem? I would have tried to debug, but I only have a Windows system with VS 2019.

ittegrat:
I've managed to compile and debug scintillua with VS 2019.
It seems to me that the 64bit crash lies in the [StringResult](https://github.com/orbitalquark/scintillua/blob/3428dfc0a871b83e8589f7ae99202fb4e3965044/LexLPeg.cxx#L649) method, as a 64bit [pointer](https://github.com/orbitalquark/scintillua/blob/3428dfc0a871b83e8589f7ae99202fb4e3965044/LexLPeg.cxx#L838) is truncated to fit into the 32bit `long lparam` parameter.
MSVC defines `int` and `long` as 32bit values, but I don't know what happens when you cross-compile with mingw.
Using `uintptr_t` seems to solve the problem, but I didn't test througly.

orbitalquark:
Thanks for the report and for looking into this! I will investigate when I have some time.

orbitalquark:
Thanks again. Fixed via https://github.com/orbitalquark/scintillua/commit/b256ddf12ef82bc6ee2089bb6bf2588874cee231

PR #55, opened 2021-09-17
Fix pointer truncation on Windows 64bit

ittegrat:
This patch should fix the issue #53


orbitalquark:
Thanks so much for looking into this and fixing!

Fixed via https://github.com/orbitalquark/scintillua/commit/b256ddf12ef82bc6ee2089bb6bf2588874cee231

PR #59, opened 2021-12-05
Fix Perl lexer number parsing

snoopy:
* Numbers can be preceeded by "0b", "0x" and "0"
* Numbers can be visually separated via "_"

See https://www.perltutorial.org/perl-numbers/ and https://perldoc.perl.org/perlnumber


orbitalquark:
Thanks! Fixed via https://github.com/orbitalquark/scintillua/commit/2a249b14904665ec4fab49589d1829f04c1ed2fb with minor reformatting and an additional check after binary numbers.

PR #60, opened 2021-12-13
Fix binary number parsing in C++.

snoopy:
Your additional fix (from 8b0e1b0c53e97d62234a4345efbf93cf6ffe4f7c and 2a249b14904665ec4fab49589d1829f04c1ed2fb) also makes sense for C++ I think.


orbitalquark:
Fixed via https://github.com/orbitalquark/scintillua/commit/4cb1464ef738a098f008d6530b776fe780b19c34 along with other affected languages.

Thanks for catching this!

PR #62, opened 2022-03-11
Add classes to Python

paaguti:
Low hanging fruit: use the class token for the class name in `class ClassName`. Code from the Java lexer.


orbitalquark:
Added via https://github.com/orbitalquark/scintillua/commit/530810c22ab92b2dbc0e0e627a5b9de3cd236475

Thanks!

PR #64, opened 2022-06-08
Update fortran.lua

hckiang:
More Fortran 90+ keywords for modern Fortran. Not sure if this is appropriate but these does make my syntax highlighting much more natural/closer to Emacs' one in Fortran 90.


orbitalquark:
Thanks! I've applied these changes via https://github.com/orbitalquark/scintillua/commit/7f7575a41426f7c258a125b4d8f2bcbf9797c56f, but moved the ISO C bindings from keywords to types.

PR #65, opened 2022-06-10
Add support for Hare language

qiu-x:
https://harelang.org


orbitalquark:
Thanks for your contribution! This lexer is in the legacy format and will need to be converted to the new one (see https://orbitalquark.github.io/scintillua/api.html#migrating-legacy-lexers). I can do this for you, but only when I have some time.

qiu-x:
I converted the lexer to the new format

orbitalquark:
Thank you! Committed via https://github.com/orbitalquark/scintillua/commit/9088723504b19f8611b66c119933a4dc7939d7b8 with some formatting changes to align with other lexers. Also fixed a bug in the `word_match()` calls to use tables.

PR #66, opened 2022-06-29
add Janet lexer

sevvie:
This was originally created for the vis repository, but it was recommended that the lexer itself be submitted upstream. This is just a copy of the clojure.lua lexer with the appropriate tokens for Janet (https://janet-lang.org)

### DIFF
--- a/lexers/janet.lua
+++ b/lexers/janet.lua
@@ -1,0 +1,178 @@
+-- Copyright 2006-2017 Mitchell mitchell.att.foicica.com. See LICENSE.
+-- Altered for Janet by sevvie Rose
+-- Janet LPeg lexer.
+
+local l = require('lexer')
+local token, word_match = l.token, l.word_match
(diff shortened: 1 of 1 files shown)


qiu-x:
Hello, it seems like this lexer is written in the legacy format. Before it can be merged it has to be updated to the new format (see https://orbitalquark.github.io/scintillua/api.html#migrating-legacy-lexers)

orbitalquark:
Thanks for your contribution. As mentioned previously, this lexer is in a legacy format and should be converted. I can do it myself, but only when I find some time.

orbitalquark:
By the way, this lexer is also incomplete, based on Janet syntax: https://janet-lang.org/docs/syntax.html

It's missing highlighting for `nil`, `true`, and `false`; numbers like `1.3e18` and `1_000_000`; long strings (backticks); and possibly shorthands.

It's quite possible the Clojure lexer is incomplete (since that's what this lexer is based on), but that is another issue.

PR #67, opened 2022-06-29
Add a lexer for .spec files used for creating RPM packages (Linux distros)

mcepl:
This is WIP still, I don’t know what’s wrong with it, but [vis](https://github.com/martanne/vis/) doesn’t show anything.

### DIFF
--- a/lexers/rpmspec.lua
+++ b/lexers/rpmspec.lua
@@ -18,11 +18,19 @@ lex:add_rule('comment', comment)
 local string = token(lexer.STRING, lexer.delimited_range('"'))
 lex:add_rule('string', string)
 
--- -- Keywords.
--- local keyword = token(lexer.KEYWORD, word_match({
---   ''
--- }, nil, true))
--- lex:add_rule('keyword', keyword)
+-- Keywords.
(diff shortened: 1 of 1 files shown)


lexers/rpmspec.lua
```diff
@@ -0,0 +1,49 @@
...
+local command = token(lexer.FUNCTION, '%' * lexer.word)
+lex:add_rule('command', command)
+
+-- -- Constants.
+-- local constant = token(lexer.CONSTANT, word_match({
+-- }, nil, true))
+-- lex:add_rule('constant', constant)
+
+-- Identifiers.
+local identifier = token(lexer.IDENTIFIER, word_match({
```
orbitalquark, line 45: These words are not highlighted any different than unrecognized/unstyled text. Do you want to make these keywords instead?

mcepl: Right

orbitalquark:
Thanks! Commited via https://github.com/orbitalquark/scintillua/commit/ecafeda70fc29602773101620ed345e10bd3325b with some styling changes and substitution for deprecated functions.

PR #70, opened 2022-07-14
Update rpmspec lexer, to make it more useable.

mcepl:
Unfortunately, this is not working, and I would like to ask for help why.

All these lines should be marked as keywords, but they are not:
```spec
Source: source.tar.gz
Source0: anothersource.tar.gz
Patch:  first.patch
Patch0: second.patch
```

### DIFF
--- a/lexers/rpmspec.lua
+++ b/lexers/rpmspec.lua
@@ -12,8 +12,8 @@ lex:add_rule('whitespace', token(lexer.WHITESPACE, lexer.space^1))
 -- Comments.
 lex:add_rule('comment', token(lexer.COMMENT, lexer.to_eol('#')))
 
--- Numbers (including versions)
-lex:add_rule('number', token(lexer.NUMBER, lexer.number + S('.+~')))
+-- -- Numbers (including versions)
+-- lex:add_rule('number', token(lexer.NUMBER, lexer.number + S('.+~')))
 
 -- Operators
 lex:add_rule('operator', token(lexer.OPERATOR, S('&<>=|')))
--- a/lexers/rpmspec.lua
+++ b/lexers/rpmspec.lua
@@ -16,7 +16,7 @@ lex:add_rule('comment', token(lexer.COMMENT, lexer.to_eol('#')))
 lex:add_rule('string', token(lexer.STRING, lexer.range('"')))
 
 -- Keywords.
-lex:add_rule('keyword', token(lexer.KEYWORD, word_match{
+lex:add_rule('keyword', token(lexer.KEYWORD, word_match({
     'Prereq', 'Summary', 'Name', 'Version', 'Packager',
     'Requires', 'Recommends', 'Suggests', 'Supplements',
     'Enhances', 'Icon', 'URL', 'Prefix', 'Packager', 'Group',
@@ -25,7 +25,7 @@ lex:add_rule('keyword', token(lexer.KEYWORD, word_match{
     'Obsoletes', 'BuildArch', 'BuildArchitectures', 'BuildRequires',
     'BuildConflicts', 'BuildPreReq', 'Conflicts', 'AutoRequires',
     'AutoReq', 'AutoReqProv', 'AutoProv', 'Epoch'
-}) + (P('Patch') + P('Source')) * R('09')^0)
+}) + (P('Patch') + P('Source')) * R('09')^0))
 
 -- Macros
 lex:add_rule('command', token(lexer.FUNCTION, '%' * lexer.word +
--- a/lexers/rpmspec.lua
+++ b/lexers/rpmspec.lua
@@ -12,6 +12,12 @@ lex:add_rule('whitespace', token(lexer.WHITESPACE, lexer.space^1))
 -- Comments.
 lex:add_rule('comment', token(lexer.COMMENT, lexer.to_eol('#')))
 
+-- Numbers (including versions)
+lex:add_rule('number', token(lexer.NUMBER, lexer.number + S('.+~')))
+
+-- Operators
+lex:add_rule('operator', token(lexer.OPERATOR, S('&<>=|')))
+
 -- Strings.
 lex:add_rule('string', token(lexer.STRING, lexer.range('"')))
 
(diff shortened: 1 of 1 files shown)


lexers/rpmspec.lua
```diff
@@ -17,15 +17,23 @@ lex:add_rule('string', token(lexer.STRING, lexer.range('"')))
...
-}))
+    'Prereq', 'Summary', 'Name', 'Version', 'Packager',
+    'Requires', 'Recommends', 'Suggests', 'Supplements',
+    'Enhances', 'Icon', 'URL', 'Prefix', 'Packager', 'Group',
+    'License', 'Release', 'BuildRoot', 'Distribution', 'Vendor',
+    'Provides', 'ExclusiveArch', 'ExcludeArch', 'ExclusiveOS',
+    'Obsoletes', 'BuildArch', 'BuildArchitectures', 'BuildRequires',
+    'BuildConflicts', 'BuildPreReq', 'Conflicts', 'AutoRequires',
+    'AutoReq', 'AutoReqProv', 'AutoProv', 'Epoch'
+}) + (P('Patch') + P('Source')) * R('09')^0)
```
orbitalquark: You are excluding these additional keywords from the `token()` call. Move the `)` from `})` to the end of the line and it should work.

mcepl: Fixed in 9d0129edba37fba708969b46c779c8764e911348.

Implied parentheses in Lua are EVIL!

lexers/rpmspec.lua
```diff
@@ -16,16 +16,24 @@ lex:add_rule('comment', token(lexer.COMMENT, lexer.to_eol('#')))
...
+    'Provides', 'ExclusiveArch', 'ExcludeArch', 'ExclusiveOS',
+    'Obsoletes', 'BuildArch', 'BuildArchitectures', 'BuildRequires',
+    'BuildConflicts', 'BuildPreReq', 'Conflicts', 'AutoRequires',
+    'AutoReq', 'AutoReqProv', 'AutoProv', 'Epoch'
+}) + (P('Patch') + P('Source')) * R('09')^0))
 
 -- Macros
-lex:add_rule('command', token(lexer.FUNCTION, '%' * lexer.word))
+lex:add_rule('command', token(lexer.FUNCTION, '%' * lexer.word +
+                                              '%{' * lexer.word * '}'))
```
orbitalquark: According to https://rpm-software-management.github.io/rpm/manual/spec.html,

`%if %{defined with_foo} && %{undefined with_bar}`

is valid. Your pattern doesn't work here. You could try `lexer.range()`, but I think maybe the insides should be highlighted since something like

`%if 0%{?fedora} > 10 || 0%{?rhel} > 7`

is also valid and you've added `fedora` and `rhel` constants.

orbitalquark (changes requested):
Maybe also add operators? It seems like those are valid too.

You may also consider adding numbers in the form of version strings, accounting for '-' and '~' (e.g. `1.17-1` and `1.1~201601`)

lexers/rpmspec.lua
mcepl: screenshot-2022-07-15_23-07-1657922283

I have to say I am bit confused about this.

1. OK, so perhaps this doesn’t work because the opening and closing brace is not on the same line, right?
```spec
%{python_expand /bin/true
$python setup.py test -v
}
```
2. Here probably the dollar sign is not in `lexer.word`
```spec
%python_expand %fdupes %{buildroot}%{$python_sitelib}
```

3. But I completely don’t understand these ones. Missing space?
```spec
%if %{defined with_foo} && %{undefined with_bar}
```
I have tried these changes, but it doesn’t seem to make any difference:

```diff
diff --git a/lexers/rpmspec.lua b/lexers/rpmspec.lua
index 6337769..ac6d219 100644
--- a/lexers/rpmspec.lua
+++ b/lexers/rpmspec.lua
@@ -5,6 +5,7 @@ local token, word_match = lexer.token, lexer.word_match
 local C, P, R, S = lpeg.C, lpeg.P, lpeg.R, lpeg.S
 
 local lex = lexer.new('rpmspec')
+local name = lexer.word * S(' $')
 
 -- Whitespace.
 lex:add_rule('whitespace', token(lexer.WHITESPACE, lexer.space^1))
@@ -12,6 +13,11 @@ lex:add_rule('whitespace', token(lexer.WHITESPACE, lexer.space^1))
 -- Comments.
 lex:add_rule('comment', token(lexer.COMMENT, lexer.to_eol('#')))
 
+-- Constants
+lex:add_rule('constant', token(lexer.CONSTANT, word_match({
+    'rhel', 'fedora', 'suse_version', 'sle_version'
+})))
+
 -- Strings.
 lex:add_rule('string', token(lexer.STRING, lexer.range('"')))
 
@@ -28,12 +34,7 @@ lex:add_rule('keyword', token(lexer.KEYWORD, word_match({
 }) + (P('Patch') + P('Source')) * R('09')^0))
 
 -- Macros
-lex:add_rule('command', token(lexer.FUNCTION, '%' * lexer.word +
-                                              '%{' * lexer.word * '}'))
-
--- Constants
-lex:add_rule('constant', token(lexer.CONSTANT, word_match({
-    'rhel', 'fedora', 'suse_version', 'sle_version'
-})))
+lex:add_rule('command', token(lexer.FUNCTION, '%' * name +
+                                              '%{' * name * '}'))
 
 return lex
```

mcepl:
Generally, you are right, this is still not finished. I will work on this a little bit more.

lexers/rpmspec.lua
orbitalquark: 1. No, it is because your pattern is only matching a single word inside `{}`. You could try `lexer.range()` to match anything inside `{}`. You'll need to add an argument to match multiple lines. However, everything between `{}` will be highlighted the same. You may want to consider making `{}` be operator characters so that anything inside can also be highlighted.
2. Correct.
3. Yes, see my answer to 1.
4. Again, see 1. If you're just trying to match words inside `{}`, with or without `$`, you won't be able to match arbitrary code inside `{}`.

PR #71, opened 2022-07-18
conversion of rest lexer to the modern style.

mcepl:
Complete don’t understand how to do M._fold function.

### DIFF
--- a/lexers/rest.lua
+++ b/lexers/rest.lua
@@ -198,35 +198,35 @@ local sphinx_levels = {
   ['#'] = 0, ['*'] = 1, ['='] = 2, ['-'] = 3, ['^'] = 4, ['"'] = 5
 }
 
--- -- Section-based folding.
--- M._fold = function(text, start_pos, start_line, start_level)
---   local folds, line_starts = {}, {}
---   for pos in (text .. '\n'):gmatch('().-\r?\n') do
---     line_starts[#line_starts + 1] = pos
---   end
---   local style_at, CONSTANT, level = lexer.style_at, lexer.CONSTANT, start_level
---   local sphinx = lexer.property_int['fold.by.sphinx.convention'] > 0
---   local FOLD_BASE = lexer.FOLD_BASE
---   local FOLD_HEADER, FOLD_BLANK = lexer.FOLD_HEADER, lexer.FOLD_BLANK
---   for i = 1, #line_starts do
---     local pos, next_pos = line_starts[i], line_starts[i + 1]
---     local c = text:sub(pos, pos)
---     local line_num = start_line + i - 1
---     folds[line_num] = level
---     if style_at[start_pos + pos - 1] == CONSTANT and c:find('^[^%w%s]') then
---       local sphinx_level = FOLD_BASE + (sphinx_levels[c] or #sphinx_levels)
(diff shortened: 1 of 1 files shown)


lexers/rest.lua
```diff
@@ -131,122 +133,109 @@ local indented_block = function(input, index)
...
-      level = (not sphinx and level or sphinx_level) + 1
-    elseif c == '\r' or c == '\n' then
-      folds[line_num] = level + FOLD_BLANK
-    end
-  end
-  return folds
-end
-
-l.property['fold.by.sphinx.convention'] = '0'
+-- -- Section-based folding.
```
mcepl: This function should be probably replaced by the series of `lexer.add_fold_point()` calls, but I have no idea how.

@orbitalquark Would you be able to help, please?

orbitalquark: You should be able to just change the name from `_fold` to `fold`. This will override the default `lexer.fold()` implementation.

lexers/rest.lua
```diff
@@ -32,23 +32,24 @@ local underline = lpeg.Cmt(starts_line(adornment), function(_, index, adm, c)
   return pos and index - #adm + pos - 1 or nil
 end)
 -- Token needs to be a predefined one in order for folder to work.
-local title = token(l.CONSTANT, overline + underline)
+lex:add_rule('title', token(lexer.CONSTANT, overline + underline))
```
orbitalquark: Be careful here because 'title' is not the first rule. (See `_rules` in the legacy lexer, which has 'literal_block' first.) Either call `lex:add_rule()` all at once later, or rearrange all of the patterns and rules to match the legacy lexer rule order.

lexers/rest.lua
mcepl: Yes, `lexer.fold()` or `lex.fold()`?

orbitalquark: `lex.fold()`. If you change `lexer.fold()` then that affects all lexers, which would be bad.

mcepl: And what about that
```lua
l.property['fold.by.sphinx.convention'] = '0'
```
`lex.property` is apparently `nil`, so my editor breaks on this line. I have commented it out for now.

orbitalquark: Did you try `lexer.property[]`? That field belongs to the `lexer` module, not a lexer object.

mcepl: Thank you.

mcepl:
@orbitalquark I think this should be ready for merge.

orbitalquark:

orbitalquark:
Once again, thanks for helping out with this. I've committed these changes in https://github.com/orbitalquark/scintillua/commit/064945527e6871fd0cbec5a9ee4b71f22ea7d635 with changes in formatting and the following:

* Moved `lex:add_style()` calls to be closer to their token definitions.
* Use style property tables in `lex:add_style()` calls.
* Consolidated `lex:add_rule()` calls to match convention of modern lexers.
* Updated `lexer.word_match()` calls to match convention of modern lexers.
* Changed `lex.fold()` to `lex:fold()` in order to fix bug in folding.

mcepl:
Thank you for clean up!

Issue #68, opened 2022-06-30
Merge vis lexers


Nomarian:
Missing lexers from vis should be migrated here instead.

- dsv
- gemini
- git-rebase
- strace
- vbscript

orbitalquark:
I chose not to include the first four lexers originally submitted here: https://github.com/orbitalquark/scintillua/pull/13

vbscript is identical to vb and is no longer needed separately: https://github.com/orbitalquark/scintillua/commit/dee7d765a0057eca938412d63b3e95742d7a5a65

orbitalquark:
I chose not to include the first four lexers originally submitted here: https://github.com/orbitalquark/scintillua/pull/13

vbscript is identical to vb and is no longer needed separately: https://github.com/orbitalquark/scintillua/commit/dee7d765a0057eca938412d63b3e95742d7a5a65

PR #72, opened 2022-07-21
Add modern rewrite of git-rebase lexer from vis.

mcepl:
Mistakenly considered included as a part of port of #13 in #68, but it has never been.

### DIFF
--- a/lexers/git-rebase.lua
+++ b/lexers/git-rebase.lua
@@ -1,0 +1,41 @@
+-- Copyright 2017-2021 Marc André Tanner
+-- git-rebase(1) LPeg lexer.
+
+local lexer = require('lexer')
+local token, word_match = lexer.token, lexer.word_match
+local P, R = lpeg.P, lpeg.R
+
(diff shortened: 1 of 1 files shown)


mcepl:
The typical filename of the file in question is `.git/rebase-merge/git-rebase-todo` (without a dot in the basename), so I am not sure how to record it in lpeg.properties.

orbitalquark:
As mentioned in #13 and #68, I opted not to include this lexer. In general, Scintillua has lexers for programming languages, markup languages, and input and configuration files. Output files are not included.

mcepl:
OK, we will have to carry this in vis alone. Oh well.

MuhammedZakir:
Just my 2 cents: IMHO, it would be nice if this repo includes commonly used stuff in addition to what you mentioned above; doing git-rebase with your favorite editor is common. :-) Another reason is that this repo is used mainly for editors. Having a common source helps.

mcepl:
I agree, but I am not willing to fight with @orbitalquark too much. It is his project, and although I may not agree with him, it is his show.

PR #73, opened 2022-07-21
Add modern rewrite of strace lexer from vis.

mcepl:
Mistakenly considered included as a part of port of #13 in #68, but it has never been.

Again, I have a problem with what to write into lpeg.properties. In vis we have

```lua
        strace = {
                detect = function(_, data)
                        return data:match("^execve%(")
                end
        },
```

but that’s probably something scintilua cannot do, right?

### DIFF
--- a/lexers/strace.lua
+++ b/lexers/strace.lua
@@ -1,0 +1,35 @@
+-- Copyright 2017-2021 Marc André Tanner. See LICENSE.
+-- strace(1) output lexer
+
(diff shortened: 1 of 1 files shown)


orbitalquark:
As mentioned in #13 and #68, I opted not to include this lexer. In general, Scintillua has lexers for programming languages, markup languages, and input and configuration files. Output files are not included.

PR #80, opened 2022-10-10
Convert Perl lexer to new format

snoopy:
I've converted the Perl lexer to the new format. Based on my quick testing it seems to work fine.
Could you please have a look as well?

### DIFF
--- a/lexers/perl.lua
+++ b/lexers/perl.lua
@@ -94,6 +94,22 @@ lex:add_rule('number', lex:tag(lexer.NUMBER, lexer.number_('_')))
 
 -- Variables.
 -- LuaFormatter off
+local builtin_var_s = '$' * (
+  lpeg.R('09') +
+  S('!"$%&\'()+,-./:;<=>?@\\]_`|~') +
+  '^' * S('ACDEFHILMNOPRSTVWX')^-1 +
+  'ARGV')
+local builtin_var_a = '@' * (
+  S('+-_F') +
+  'ARGV' + 'INC' + 'ISA')
+local builtin_var_h = '%' * (
+  S('+-!') +
+  '^' * S('H')^-1 +
+  'ENV' + 'INC' + 'SIG')
+lex:add_rule('variable_builtin', lex:tag(lexer.VARIABLE_BUILTIN,
+  builtin_var_s +
+  builtin_var_a +
+  builtin_var_h))
 local special_var = '$' * (
   '^' * S('ADEFHILMOPSTWX')^-1 +
   S('\\"[]\'&`+*.,;=%~?@<>(|/!-') +
--- a/lexers/perl.lua
+++ b/lexers/perl.lua
@@ -10,7 +10,7 @@ local lex = lexer.new(...)
 lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD)))
 
 -- Markers.
-lex:add_rule('marker', lex:tag(lexer.COMMENT, lex:word_match(lexer.COMMENT) * lexer.any^0))
+lex:add_rule('marker', lex:tag(lexer.COMMENT, lexer.word_match('__DATA__ __END__') * lexer.any^0))
 
 -- Strings.
 local delimiter_matches = {['('] = ')', ['['] = ']', ['{'] = '}', ['<'] = '>'}
@@ -143,10 +143,6 @@ lex:set_word_list(lexer.FUNCTION_BUILTIN, {
   'sysseek', 'system', 'syswrite', 'telldir', 'tell', 'tied', 'tie', 'time', 'times', 'truncate',
   'ucfirst', 'uc', 'umask', 'undef', 'unlink', 'unpack', 'unshift', 'untie', 'utime', 'values',
   'vec', 'wait', 'waitpid', 'wantarray', 'warn', 'write'
-})
-
-lex:set_word_list(lexer.COMMENT, {
-  '__DATA__', '__END__'
 })
 
 lexer.property['scintillua.comment'] = '#'
--- a/lexers/perl.lua
+++ b/lexers/perl.lua
@@ -81,6 +81,9 @@ local func = lex:tag(lexer.FUNCTION, lexer.word)
 local method = B('->') * lex:tag(lexer.FUNCTION_METHOD, lexer.word)
 lex:add_rule('function', (method + func) * #(lexer.space^0 * '('))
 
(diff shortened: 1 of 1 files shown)


orbitalquark:
Thank you! I will take a look when I have some time. Sorry for the delay.

lexers/perl.lua
```diff
@@ -1,52 +1,16 @@
...
-  'rename', 'reset', 'reverse', 'rewinddir', 'rindex', 'rmdir', 'scalar', 'seekdir', 'seek',
-  'select', 'semctl', 'semget', 'semop', 'send', 'setgrent', 'sethostent', 'setnetent', 'setpgrp',
-  'setpriority', 'setprotoent', 'setpwent', 'setservent', 'setsockopt', 'shift', 'shmctl', 'shmget',
-  'shmread', 'shmwrite', 'shutdown', 'sin', 'sleep', 'socket', 'socketpair', 'sort', 'splice',
-  'split', 'sprintf', 'sqrt', 'srand', 'stat', 'study', 'substr', 'symlink', 'syscall', 'sysread',
-  'sysseek', 'system', 'syswrite', 'telldir', 'tell', 'tied', 'tie', 'time', 'times', 'truncate',
-  'ucfirst', 'uc', 'umask', 'undef', 'unlink', 'unpack', 'unshift', 'untie', 'utime', 'values',
-  'vec', 'wait', 'waitpid', 'wantarray', 'warn', 'write'
-}))
+lex:add_rule('marker', lex:tag(lexer.COMMENT, lex:word_match(lexer.COMMENT) * lexer.any^0))
```
orbitalquark: I would switch back to `lexer.word_match()` here because comment words are quite static.

lexers/perl.lua
```diff
@@ -102,25 +66,31 @@ local lit_str = 'q' * P('q')^-1 * literal_delimited
...
   sq_str + dq_str + cmd_str + heredoc + lit_str + lit_array + lit_cmd + lit_tr)
 local regex_str = lexer.after_set('-<>+*!~\\=%&|^?:;([{', lexer.range('/', true) * S('imosx')^0)
 local lit_regex = 'qr' * literal_delimited * S('imosx')^0
 local lit_match = 'm' * literal_delimited * S('cgimosx')^0
 local lit_sub = 's' * literal_delimited2 * S('ecgimosx')^0
-local regex = token(lexer.REGEX, regex_str + lit_regex + lit_match + lit_sub)
+local regex = lex:tag(lexer.REGEX, regex_str + lit_regex + lit_match + lit_sub)
 lex:add_rule('string', string + regex)
 
+-- Functions.
```
orbitalquark, line 78: What was the reason for switching the rule order here? Is there something about strings that prevents them from matching after functions/methods? (I do not know much about Perl.)

orbitalquark (changes requested):
This looks good for the most part. Thank you for taking the time to migrate this lexer! Does Perl have any built-in constants worth adding? This would be a good opportunity to do so. Otherwise, no problem.

lexers/perl.lua
snoopy: During testing the old order broke regex strings. The easiest fix was adjusting the rule order.
Perl being Perl it still might break in unexpected ways.

snoopy:
> [...] Does Perl have any built-in constants worth adding? This would be a good opportunity to do so. Otherwise, no problem.

Not that I know of unless you count special vars (https://perldoc.perl.org/variables). They are not really constant though.

orbitalquark:
Can we use `lexer.VARIABLE_BUILTIN` for those? I think that will be useful. Maybe not all of them, but some useful subset, mainly so that the word list exists and other users can extend it to their liking.

lexers/perl.lua
snoopy: You mean like this? 
```lua
lex:add_rule('marker', lex:tag(lexer.COMMENT, lex:word_match('__DATA__ __END__')))

Comment (orbitalquark) [reply lexers/perl.lua]: No, just straight up `lexer.word_match('__DATA__ __END__')`. The updated JSON lexer uses it: https://github.com/orbitalquark/scintillua/blob/aaabb16898389b0e7a130ca5f31db420cfa4164a/lexers/json.lua#L16

Comment (snoopy) [reply lexers/perl.lua]:
```lua
lex:add_rule('marker', lex:tag(lexer.KEYWORD, lexer.word_match('__DATA__ __END__')))
```

This actually does not style everything after `__DATA__` as comment which might be nicer.

orbitalquark: It would be nicer, but rules that span multiple lines are quite brittle in my experience.

snoopy:
Variables are a bit spotty. LMK if there are any obvious mistakes.

snoopy:
Something is still wrong with the function/string rules. I'll try to figure it out.

snoopy:
Functions should work better now. The variables are still having problems where some of them don't get detected properly. The rules are probably not quite right.

What kind of changes should be made to the marker rule?

orbitalquark:
The marker rule should still mark `__DATA__` or `__END__` followed by anything after as a comment, just like before.

lexers/perl.lua
```diff
@@ -132,15 +102,143 @@ local special_var = '$' * (
   lexer.digit^1)
 -- LuaFormatter on
 local plain_var = ('$#' + S('$@%')) * P('$')^0 * lexer.word + '$#'
-lex:add_rule('variable', token(lexer.VARIABLE, special_var + plain_var))
+lex:add_rule('variable_builtin', lex:tag(lexer.VARIABLE_BUILTIN, lex:word_match(lexer.VARIABLE_BUILTIN)))
```
orbitalquark: Built-in variables should not all be grouped into a single word match. Scintillua looks for the longest sequence of characters matching letters and all special characters used inside a word list and then does the lookup. This is probably the cause of not detecting some variables properly.

Instead, something like `plain_var` above should be done for grouping variables into subsets. For example, ``'$' * (lpeg.R('19') + S('!"$%&\'()+,-./:;<=>?@\\]_`|~') + '^' * S('ACDEFHILMNOPRSTVWX')^-1 + 'ARGV'`` will cover most of the '$' variables. You can do something similar for '%' and '@' variables.

snoopy:
Thanks for the pointers. Seems to work fine now.

orbitalquark:
Thank you! Committed with some minor formatting changes via https://github.com/orbitalquark/scintillua/commit/ce13dedf4a0d7427e3063271d7c8ca89c0530f16

PR #81, opened 2022-11-18
Convert autoit lexer

snoopy:
Converts the Autoit lexer to the new format.

### DIFF
--- a/lexers/autoit.lua
+++ b/lexers/autoit.lua
@@ -2,26 +2,58 @@
 -- AutoIt LPeg lexer.
 -- Contributed by Jeff Stone.
 
-local lexer = require('lexer')
-local token, word_match = lexer.token, lexer.word_match
-local P, S = lpeg.P, lpeg.S
+local lexer = lexer
+local P, S, B = lpeg.P, lpeg.S, lpeg.B
 
-local lex = lexer.new('autoit')
-
--- Whitespace.
-lex:add_rule('whitespace', token(lexer.WHITESPACE, lexer.space^1))
+local lex = lexer.new(...)
 
 -- Keywords.
-lex:add_rule('keyword', token(lexer.KEYWORD, word_match({
+lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD, true)))
+
+-- Functions.
+local builtin_func = lex:tag(lexer.FUNCTION_BUILTIN, lex:word_match(lexer.FUNCTION_BUILTIN, true))
+local func = lex:tag(lexer.FUNCTION, lexer.word)
+local method = B('.') * lex:tag(lexer.FUNCTION_METHOD, lexer.word)
+lex:add_rule('function', (builtin_func + method + func) * #(lexer.space^0 * '('))
+
+-- Identifiers.
+lex:add_rule('identifier', lex:tag(lexer.IDENTIFIER, lexer.word))
+
+-- Comments.
+local line_comment = lexer.to_eol(';')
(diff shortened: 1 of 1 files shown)


lexers/autoit.lua
```diff
@@ -2,26 +2,59 @@
...
 
-local lex = lexer.new('autoit')
-
--- Whitespace.
-lex:add_rule('whitespace', token(lexer.WHITESPACE, lexer.space^1))
+local lex = lexer.new(...)
 
 -- Keywords.
-lex:add_rule('keyword', token(lexer.KEYWORD, word_match({
+lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD)))
```
orbitalquark: Keywords were case-insensitive in the previous lexer. If they are still case-insensitive, please call `lex:word_match(lexer.KEYWORD, true)`. The same idea applies to other word lists.

orbitalquark (changes requested):
Thanks for your PR! I really appreciate it :)

lexers/autoit.lua
snoopy: Thanks, works as expected.

lexers/autoit.lua
```diff
@@ -2,26 +2,59 @@
...
+local lex = lexer.new(...)
 
 -- Keywords.
-lex:add_rule('keyword', token(lexer.KEYWORD, word_match({
+lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD, true)))
+
+-- Functions.
+lex:add_rule('function_builtin',
+  lex:tag(lexer.FUNCTION_BUILTIN, lex:word_match(lexer.FUNCTION_BUILTIN, true)) *
+    #(lexer.space^0 * P('(')^-1))
```
orbitalquark: `#(lexer.space^0 * P('(')^-1)` doesn't do anything here, so I would either:

* Remove the `^-1` if AutoIt allows variables/constants to use built-in function names. (Basically, '(' is required for this to be recognized as a built-in function.)
* Remove the entire expression if these words can only be interpreted as built-in functions and they don't require a trailing '(' to be used (like how functions are first-class values in Lua).

lexers/autoit.lua
```diff
@@ -2,26 +2,59 @@
...
+local func = lex:tag(lexer.FUNCTION, lexer.word)
+lex:add_rule('function', S('_')^0 * func * #(lexer.space^0 * '('))
```
orbitalquark: `S('_')^0` shouldn't be needed because `lexer.word` recognizes '_' as word characters.

orbitalquark: I would perhaps combine these two rules into a single `'function'` rule:
```
local builtin_func = lex:tag(lexer.FUNCTION_BUILTIN, ...)
local func = lex:tag(lexer.FUNCTION, ...)
lex:add_rule('function', builtin_func + func)
```
This is similar to other migrated lexers.

orbitalquark (changes requested):
This is looking good. I think we're almost there.

lexers/autoit.lua
snoopy: Since both languages are very similar I ended up reusing the function rule from AutoHotkey here.

lexers/autoit.lua
snoopy: Good point.

snoopy: Done, LMK if this is okay.

orbitalquark:
Committed via https://github.com/orbitalquark/scintillua/commit/ba11af48d396b9fc11cca026d441ce2ed17d51cb. Thanks so much!

PR #82, opened 2022-11-18
Add AutoHotkey lexer

snoopy:
Adds a lexer for AutoHotkey (https://en.wikipedia.org/wiki/AutoHotkey) based on the AutoIt lexer.

Is there a way to enforce both a minimum and maximum amount of characters via `patt^n` and `patt^-n`? This would be for an F-key rule (match F1-F12 but don't match F123 or F by itself).

Is it possible for word_match to be case insensitive? Currently it does not seem to work.


orbitalquark:
I can't think of a way to do that. The CSS lexer recognizes `#rgb` and `#rrggbb` using `xdigit * xdigit * xdigit * (xdigit * xdigit * xdigit)^-1`, so perhaps you can do something similar.

Are there any file extensions for AutoHotkey scripts? You can add them to *lexer.lua*'s `extensions` table in the `detect()` function. (Or I can do so if you let me know.)

`word_match()` accepts a second argument for case-insensitivity. It works at least for the vb lexer.

snoopy:
Many thanks for your input!
I noticed you even used the case-insensitivity in your latest commit. Sorry for not paying attention to that.

lexers/autohotkey.lua
```diff
@@ -0,0 +1,665 @@
...
+lex:add_rule('variable_builtin', lex:tag(lexer.VARIABLE_BUILTIN, 'A_' * lex:word_match(lexer.VARIABLE_BUILTIN, true)))
+
+-- Constants.
+lex:add_rule('constant', lex:tag(lexer.CONSTANT_BUILTIN, S('fF') * lexer.digit * (lexer.digit)^-1
+    + lex:word_match(lexer.CONSTANT_BUILTIN, true)))
+
+-- Functions.
+lex:add_rule('function_builtin',
+  lex:tag(lexer.FUNCTION_BUILTIN, lex:word_match(lexer.FUNCTION_BUILTIN, true)) *
+    #(lexer.space^0 * P('(')^-1))
```
orbitalquark: I made a comment in your AutoIt lexer about the `#(lexer.space^0 * P('(')^-1)` component, and the same idea applies.

lexers/autohotkey.lua
```diff
@@ -0,0 +1,665 @@
...
+local func = lex:tag(lexer.FUNCTION, lexer.word)
+local method = lpeg.B('.') * lex:tag(lexer.FUNCTION_METHOD, lexer.word)
+lex:add_rule('function', (method + func) * #(lexer.space^0 * '('))
```
orbitalquark: Could we perhaps combine this rule with the one above as I suggested in the AutoIt PR?
As a side note, I like the `#(lexer.space^0 * '(')` here, as it clearly requires a '(' to identify a function or method name.

lexers/autohotkey.lua
```diff
@@ -0,0 +1,665 @@
...
+
+-- Operators.
+lex:add_rule('operator', lex:tag(lexer.OPERATOR, S('~+-^*/&<>=?:()[]{}')))
+
+lex:set_word_list(lexer.KEYWORD, {
+  'as', 'and', 'contains', 'false', 'in', 'is', 'IsSet', 'not', 'or', 'super', 'true', 'unset',
+  'Break', 'Catch', 'Continue', 'Else', 'Finally', 'For', 'Global', 'Goto', 'If', 'Local', 'Loop', 'Return', 'Static', 'Throw', 'Try', 'Until', 'While'
+})
+
+lex:set_word_list(lexer.FUNCTION_BUILTIN, {
```
orbitalquark, line 51: When I commit this, is it okay if I have my Lua code formatter format it like your keyword list? If you want to separate words into sections, you can use line comments like in https://github.com/orbitalquark/scintillua/blob/3559ca9a7859a69385f94ce596233c9a1d582fc2/lexers/ansi_c.lua#L98

orbitalquark (changes requested):
Almost there. Just a couple things to address that are related to the AutoIt feedback I gave.

lexers/autohotkey.lua
snoopy: Absolutely, in fact I was relying on you to format this properly. It's only in this format because that's how it ended up after I pulled the data from the documentation and I don't currently have a lua formatter set up.

lexers/autohotkey.lua
snoopy: Done, LMK if this is okay.

lexers/autohotkey.lua
snoopy: Done, LMK if this is okay.

orbitalquark:
Committed via https://github.com/orbitalquark/scintillua/commit/5fa357a224f6186e0b0198a8e20ee55066496b58. Thank you!

PR #83, opened 2022-11-19
Convert Rust lexer

snoopy:
I made some slight changes. Mainly to deal with Rust's disastrous decision to use `'` as an operator.

### DIFF
--- a/lexers/rust.lua
+++ b/lexers/rust.lua
@@ -11,7 +11,7 @@ local lex = lexer.new(...)
 lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD)))
 
 -- Library types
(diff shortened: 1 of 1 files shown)


orbitalquark:
Thanks, this is a good start. Before committing it, I'd really like to have support for highlighting built-in constants, functions, etc., as well as support for highlighting arbitrary functions and methods (like the other migrated lexers). I don't know anything about Rust, so please use your best judgement.

If you're not feeling up for it, no problem. Just please submit a separate fix for the existing legacy lexer and we'll punt the migration to some other time.

snoopy:
I've split it up into several commits. The first one should be only the conversion itself. All the commits after that try to improve upon the base lexer.

orbitalquark:
Committed as https://github.com/orbitalquark/scintillua/commit/6e9c610eef727a413df61baf970a47ddf6cf4228

Thank you! This looks great :)

Issue #76, opened 2022-09-05
Update legacy lexers


orbitalquark:
Scintillua 6.0 considers lexers from Scintillua 5.x to be legacy lexers. While legacy lexers should still function properly, they ought to be migrated using this [migration guide][].

[migration guide]: https://orbitalquark.github.io/scintillua/api.html#migrating-legacy-lexers

In addition to the steps outlined in the migration guide, programming language lexers should ideally distinguish between functions, builtin functions, and methods. They should also distinguish between constants and builtin constants, and variables and builtin variables. For example, the Lua lexer does so [here](https://github.com/orbitalquark/scintillua/blob/f0ce9bb58ec04a8747bfbb05e85500dd629d804f/lexers/lua.lua#L13) and [here](https://github.com/orbitalquark/scintillua/blob/f0ce9bb58ec04a8747bfbb05e85500dd629d804f/lexers/lua.lua#L23). The Makefile lexer does so [here](https://github.com/orbitalquark/scintillua/blob/f0ce9bb58ec04a8747bfbb05e85500dd629d804f/lexers/makefile.lua#L26).

Lexers that use custom tokens should try and pick from the updated list of [tags][]. If that is not possible/reasonable, it may be worth considering adding to that list. Otherwise, it should remain a custom tag name and a style should be set for it for all themes in *themes/*. That will signal that applications will need to add styling for it.

The following unchecked Scintillua lexers still need to be migrated:

- [ ] Actionscript
- [ ] Ada
- [ ] ANTLR
- [ ] APDL
- [ ] APL
- [ ] Applescript
- [x] ASM (NASM)
- [x] ASP
- [x] AutoIt
- [x] AWK
- [ ] Batch
- [x] BibTeX
- [ ] Boo
- [x] C
- [x] C++
- [ ] C#
- [x] ChucK
- [ ] Clojure
- [x] CMake
- [ ] Coffeescript
- [ ] ConTeXt
- [x] CSS
- [x] CUDA
- [x] D
- [ ] Dart
- [x] Desktop Entry
- [x] Diff
- [x] Django
- [x] Dockerfile
- [ ] Dot
- [ ] Eiffel
- [ ] Elixir
- [ ] Elm
- [ ] Erlang
- [ ] F#
- [ ] Fantom
- [ ] Faust
- [ ] Fennel
- [ ] Fish
- [ ] Forth
- [ ] Fortran
- [x] fstab
- [ ] GAP
- [x] Gemini
[...]

PR #84, opened 2022-12-01
Replace `man` lexer by improved `troff` one

Eolien55:
As the title suggests, I wrote an improved version of the `man` lexer, while still retaining most of its features. `-man` headers highlighting have been kept, and the rest is way more generic (strings, all-caps/all-lowercase macros, etc.). It should also be noted that this lexer incorporates `groff`-specific elements, such as `while`s and other structured programming.

What's been added :
- generic macros (before, only a limited set of `-man` macros were recognized)
- strings (surrounded by `"` ; a bit buggy when the string is stopped by a newline rather than a matching `"`)
- escape pattern (ie `\`) recognition
- more file extensions is recognized as troff

What could be improved :
- Some macro sets (`-mom` and `-mandoc`) have a special syntax for some macros. Support could potentially added to support those.


lexers/troff.lua
```diff
@@ -0,0 +1,33 @@
+-- Copyright 2015-2017 David B. Lamkins <david@lamkins.net>. See LICENSE.
+-- man/roff LPeg lexer.
+
+local lexer = require('lexer')
+local token, word_match = l.token, l.word_match
```
orbitalquark: `l` is not defined here. I think you mean `lexer` for this and all other instances in the file.

orbitalquark (changes requested):
Thanks for your submission! As written, it is considered a "legacy lexer". Before I can accept it, it'll have to be migrated (see https://github.com/orbitalquark/scintillua/issues/76). Since this is a small lexer, it shouldn't be too much work.

lexers/troff.lua
Eolien55: This is fixed, and another typo too (`M` was returned instead of `lex`)

Eolien55:
I've migrated the lexer as you asked; is the result acceptable?

orbitalquark:
This looks better. I'll have a more thorough look when I have some time and get back to you. Thanks!

lexers/troff.lua
```diff
@@ -0,0 +1,30 @@
...
+local P, R, S = lpeg.P, lpeg.R, lpeg.S
+
+local lex = lexer.new(...)
+
+-- Registers and groff's structured programming.
+lex:add_rule('keywords', lex:tag(lexer.KEYWORD, (lexer.starts_line('.') * lexer.space^0 * (P('while') + P('break') + P('continue') + P('nr') + P('rr') + P('rnn') + P('aln') + P('\\}'))) + P('\\{'))
+
+-- Markup.
+lex:add_rule('escape_sequences', lex:tag(lexer.VARIABLE,
+	P('\\') * ((P('s') * S('+-')) + S('*fgmnYV') + P("")) *
```
orbitalquark: The `P("")` is not needed.

lexers/troff.lua
```diff
@@ -0,0 +1,30 @@
+-- Copyright 2015-2017 David B. Lamkins <david@lamkins.net>. See LICENSE.
+-- man/roff LPeg lexer.
+
+local lexer = lexer
+local P, R, S = lpeg.P, lpeg.R, lpeg.S
+
+local lex = lexer.new(...)
+
+-- Registers and groff's structured programming.
+lex:add_rule('keywords', lex:tag(lexer.KEYWORD, (lexer.starts_line('.') * lexer.space^0 * (P('while') + P('break') + P('continue') + P('nr') + P('rr') + P('rnn') + P('aln') + P('\\}'))) + P('\\{'))
```
orbitalquark: There are syntax errors throughout this file. ')' are missing.

lexers/troff.lua
```diff
@@ -0,0 +1,30 @@
...
+
+-- Markup.
+lex:add_rule('escape_sequences', lex:tag(lexer.VARIABLE,
+	P('\\') * ((P('s') * S('+-')) + S('*fgmnYV') + P("")) *
+	(P('(') * 2 + P('[') * (lexer.any-P(']'))^0 * P(']') + 1))
+
+lex:add_rule('headings', lex:tag(lexer.NUMBER, P('.') * (S('STN') * P('H')) * (lexer.space + P("\n")) * lexer.nonnewline^0)
+lex:add_rule('man_alignment', lex:tag(lexer.KEYWORD, lexer.starts_line('.'), (P('br') + P('DS') + P('RS') + P('RE') + P('PD')) * (lexer.space + P("\n")))
```
orbitalquark: Your `lex:tag()` call has an additional parameter after `lexer.stars_line('.')` that will be ignored. Perhaps you meant to use a `*` instead of a `,`?

lexers/troff.lua
```diff
@@ -0,0 +1,30 @@
...
+
+-- Registers and groff's structured programming.
+lex:add_rule('keywords', lex:tag(lexer.KEYWORD, (lexer.starts_line('.') * lexer.space^0 * (P('while') + P('break') + P('continue') + P('nr') + P('rr') + P('rnn') + P('aln') + P('\\}'))) + P('\\{'))
+
+-- Markup.
+lex:add_rule('escape_sequences', lex:tag(lexer.VARIABLE,
+	P('\\') * ((P('s') * S('+-')) + S('*fgmnYV') + P("")) *
+	(P('(') * 2 + P('[') * (lexer.any-P(']'))^0 * P(']') + 1))
+
+lex:add_rule('headings', lex:tag(lexer.NUMBER, P('.') * (S('STN') * P('H')) * (lexer.space + P("\n")) * lexer.nonnewline^0)
```
orbitalquark: The `P("\n")` is not necessary because `lexer.space` includes it.

lexers/troff.lua
```diff
@@ -0,0 +1,30 @@
...
+lex:add_rule('man_alignment', lex:tag(lexer.KEYWORD, lexer.starts_line('.'), (P('br') + P('DS') + P('RS') + P('RE') + P('PD')) * (lexer.space + P("\n")))
+lex:add_rule('font', lex:tag(lexer.VARIABLE, (P('.B') * P('R')^-1 + P('.I') * S('PR')^-1 + P('.PP')) * (lexer.space + P("\n")))
```
orbitalquark: Again, unnecessary `P("\n")`.

lexers/troff.lua
```diff
@@ -0,0 +1,30 @@
...
+
+-- Lowercase troff macros are plain macros (like .so or .nr).
+lex:add_rule('troff_plain_macros', lex:tag(lexer.VARIABLE, lexer.starts_line('.') * lexer.space^0 * lexer.lower^1)
+lex:add_rule('any_macro', lex:tag(lexer.PREPROCESSOR, lexer.starts_line('.') * lexer.space^0 * (lexer.any-lexer.space)^0)
+lex:add_rule('comment', lex:tag(lexer.COMMENT, (lexer.starts_line('.\\"') + P('\\"') + P('\\#')) * lexer.nonnewline^0)
+lex:add_rule('string', lex:tag(lexer.STRING, P('"') * (lexer.nonnewline-P('"'))^0 * (P('"') + (P("\n"))))
```
orbitalquark: You should be able to reduce this to `lexer.range('"', true, false)` since it appears to be a double-quoted range that can only span a single line and does not have escapes.

lexers/troff.lua
```diff
@@ -0,0 +1,30 @@
...
+
+-- Usually used by eqn, and mandoc in some way.
+lex:add_rule('in_dollars', lex:tag(lexer.EMBEDDED, P('$') * (lexer.any - P('$'))^0 * P('$'))
```
orbitalquark: Can this be reduced to `lexer.range('$')`?

lexers/troff.lua
```diff
@@ -0,0 +1,30 @@
...
+
+local lex = lexer.new(...)
+
+-- Registers and groff's structured programming.
+lex:add_rule('keywords', lex:tag(lexer.KEYWORD, (lexer.starts_line('.') * lexer.space^0 * (P('while') + P('break') + P('continue') + P('nr') + P('rr') + P('rnn') + P('aln') + P('\\}'))) + P('\\{'))
+
+-- Markup.
+lex:add_rule('escape_sequences', lex:tag(lexer.VARIABLE,
+	P('\\') * ((P('s') * S('+-')) + S('*fgmnYV') + P("")) *
+	(P('(') * 2 + P('[') * (lexer.any-P(']'))^0 * P(']') + 1))
```
orbitalquark: Can this be reduced to `lexer.range('[', ']')`?

orbitalquark (changes requested):
I am sorry for the delayed response and I do appreciate your patience. I have some more comments to consider on this change.

lexers/troff.lua
Eolien55: Yes, in fact I had done it, in my local `vis` lexer, but I forgot to include it in the scintillua one

Eolien55: And my implementation was very faulty, considering *roff does allow multiple line (with `\` and escapes)

lexers/troff.lua
Eolien55: I believe `eqn` has no knowledge of escapes, at least for Plan 9's and groff's. So it'd be `lexer.range('$', false, false)`

lexers/troff.lua
Eolien55: I think this range decleration doesn't allow escapes, so yes

Eolien55:
I believe I've gone over all your change requests

lexers/troff.lua
```diff
@@ -7,24 +7,24 @@ local P, R, S = lpeg.P, lpeg.R, lpeg.S
 local lex = lexer.new(...)
 
 -- Registers and groff's structured programming.
-lex:add_rule('keywords', lex:tag(lexer.KEYWORD, (lexer.starts_line('.') * lexer.space^0 * (P('while') + P('break') + P('continue') + P('nr') + P('rr') + P('rnn') + P('aln') + P('\\}'))) + P('\\{'))
+lex:add_rule('keywords', lex:tag(lexer.KEYWORD, (lexer.starts_line('.') * lexer.space^0 * (P('while') + P('break') + P('continue') + P('nr') + P('rr') + P('rnn') + P('aln') + P('\\}'))) + P('\\{')))
 
 -- Markup.
 lex:add_rule('escape_sequences', lex:tag(lexer.VARIABLE,
-	P('\\') * ((P('s') * S('+-')) + S('*fgmnYV') + P("")) *
-	(P('(') * 2 + P('[') * (lexer.any-P(']'))^0 * P(']') + 1))
```
orbitalquark: The original man lexer appeared to use `S('BIPR')` instead of `1`. By removing `+ 1`, sequences like `\fB` no longer highlight. If troff shouldn't handle this sequence, then it should not replace man, but exist alongside it.

Please make sure you're testing your changes before submitting them for review. Otherwise this process is going to go a lot more slowly.

orbitalquark (changes requested):
I'm beginning to wonder if this lexer should exist alongside the man lexer instead of replacing it. See my most recent comment. If you can get it to lex everything man does and more, then we can make a case for replacement.

lexers/troff.lua
Eolien55: `man` macros are troff syntax + special purpose macros. I have just generalized the highlighting from `man` specific macros to `troff` syntax (by adding strings, comments, more escape sequences, while retaining the macros specific to `man` (because man is extremely used and because those macros are shared by other macro sets, such as `ms`)). I did not intend to delete the +1 and do not really understand how I did, but the +1 is very important for the escape sequences to work correctly.

lexers/troff.lua
Eolien55: It is needed because sequences such as `\(em` or `\[u00E0]` need to be highlighted. But I think it is cleaner when implemented as `^-1` around `(P('s') * S('+-')) + S('*fgmnYV')`

Eolien55:
I believe the lexer works correctly, and is complete.

> If you can get it to lex everything man does and more

I have this test file, and my lexer effectively lexes everything man does, and more.

```groff
.NH 1 \" ms's way of declaring headings
A \fBHeading\fP
.so included.1
.PP
Wonderful\*[SUP]Mom macro set's way of making supertext\*[SUPX]
.PS
\" I don't know how pic works
.PE

\[some random escaped text]

\s+(88
.HEADING 1 "First-level heading" \" Mom macro's headings
```

As a sidenote, I find it quite funny that even Github doesn't highlight troff code correctly.

orbitalquark:
Added via https://github.com/orbitalquark/scintillua/commit/369ee2bb1b6e8c6a613c45e1ee88779e28060e85. Thanks for your time and patience!

PR #86, opened 2023-02-03
writeln is not a keyword

ghost:
read, write, readonly, writeonly are keywords but not writeln.


orbitalquark:
Thanks! I will apply this.

orbitalquark:
Fixed via 
https://github.com/orbitalquark/scintillua/commit/2181945f83beef745044ad6ab05f41395485ea3a

PR #94, opened 2023-07-28
Improve Hare lexer

YerinAlexey:
* Add missing keywords and operators; remove `char`, which is no longer a type
* Correctly define octal literals (`0o1234` instead of `01234`) and add binary literals
* Highlight number suffixes such as `0z` or `3.14f32`
* Highlight `size` both as a type and as `size(T)` expression
* Add missing rune literals
* Rename at_rule to annotation and remove limitations on their names
* Highlight error assertion operator with error style, similar to hare.vim

### DIFF
--- a/lexers/hare.lua
+++ b/lexers/hare.lua
@@ -5,7 +5,7 @@
 
 local lexer = require('lexer')
 local token, word_match = lexer.token, lexer.word_match
-local P, S = lpeg.P, lpeg.S
+local P, R, S = lpeg.P, lpeg.R, lpeg.S
 
 local lex = lexer.new('hare')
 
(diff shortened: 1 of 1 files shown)


orbitalquark:
Is this ready for review?

YerinAlexey:
> Is this ready for review?

Yes

lexers/hare.lua
```diff
@@ -14,47 +14,57 @@ lex:add_rule('whitespace', token(lexer.WHITESPACE, lexer.space^1))
...
 
 -- Functions.
+local size_builtin = 'size' * lexer.space^0 * #P('(')
 lex:add_rule('function', token(lexer.FUNCTION, word_match{
-  'len', 'alloc', 'free', 'assert', 'abort', 'size', 'append', 'insert', 'delete', 'vastart',
-  'vaarg', 'vaend'
-}))
+  'abort', 'align', 'alloc', 'append', 'assert', 'delete', 'free', 'insert', 'len', 'offset',
+  'vaarg', 'vaend', 'vastart'
+} + size_builtin))
```
orbitalquark: Unless Hare has first-class functions, may I suggest including `'size'` in the list and tack on `#(lexer.space^0 * '(')` after `word_match`? Otherwise, you should have `#(lexer.space^0 * '(')` so you don't include whitespace in the match.

lexers/hare.lua
```diff
@@ -14,47 +14,57 @@ lex:add_rule('whitespace', token(lexer.WHITESPACE, lexer.space^1))
...
 
 -- Comments.
 lex:add_rule('comment', token(lexer.COMMENT, lexer.to_eol('//')))
 
 -- Numbers.
-lex:add_rule('number', token(lexer.NUMBER, lexer.number))
+local literal_suffix = word_match{
+	"i", "u", "z", "i8", "i16", "i32", "i64", "u8", "u16", "u32", "u64", "f32", "f64",
+}
+local bin_num = '0b' * R('01')^1
```
orbitalquark: Would it be possible to use `lexer.bin_num` here?

lexers/hare.lua
```diff
@@ -14,47 +14,57 @@ lex:add_rule('whitespace', token(lexer.WHITESPACE, lexer.space^1))
...
+local oct_num = '0o' * R('07')^1
```
orbitalquark: Is an upper-case 'o' allowed here? Also, you might want to tack on a `* -lexer.xdigit` suffix, or non-octal numbers will match immediately after this, resulting in an apparently incorrect highlight.

orbitalquark (changes requested):

lexers/hare.lua
YerinAlexey: > Unless Hare has first-class functions, may I suggest including `'size'` in the list and tack on `#(lexer.space^0 * '(')` after `word_match`?

Requiring '(' after builtins will result in less than ideal highlighting of cases like

```
let len = 42;
```

Here `len` would be highlighted as a regular identifier because it's not followed by parenthesis, but it's actually a hard keyword and can't be used in this context.

> Otherwise, you should have `#(lexer.space^0 * '(')` so you don't include whitespace in the match.

Thanks for the suggestion, I'll add this

lexers/hare.lua
YerinAlexey: Actually, no. `lexer.bin_num` seems to detect both `0b...` and `0B...`, while Hare only supports the former. This also means that `lexer.hex_num` should be replaced by explicit `'0x' * lexer.xdigit^1`

lexers/hare.lua
YerinAlexey: > Is an upper-case 'o' allowed here?

No, just like 'b' and 'x'

> Also, you might want to tack on a `* -lexer.xdigit` suffix, or non-octal numbers will match immediately after this, resulting in an apparently incorrect highlight.

Thanks for the suggestion, fixed it

YerinAlexey: > Is an upper-case 'o' allowed here?

No, just like 'b' and 'x'

> Also, you might want to tack on a `* -lexer.xdigit` suffix, or non-octal numbers will match immediately after this, resulting in an apparently incorrect highlight.

Thanks for the suggestion, fixed it

orbitalquark:
Thanks for your contribution! Committed via https://github.com/orbitalquark/scintillua/commit/0c8d4ad957bc8e4923396d50e18307b866069e03

PR #100, opened 2023-08-31
Update Hare lexer

YerinAlexey:


orbitalquark:
I had some time to migrate the Hare lexer from legacy format, and incorporated your PR changes here: https://github.com/orbitalquark/scintillua/commit/f426182719424b27c8ce540fa3c32bcb9059d101.

An exception was leaving 'union' as a type instead of a keyword. According to the Hare spec, it seems more like a type than a keyword to me. Let me know if you strongly disagree and I'll change it too.

YerinAlexey:
> I had some time to migrate the Hare lexer from legacy format, and incorporated your PR changes here: [f426182](https://github.com/orbitalquark/scintillua/commit/f426182719424b27c8ce540fa3c32bcb9059d101).

Nice!

I'm a bit skeptical of highlighting regular functions. Can it be configured to only highlight builtins?

Also, `nullable` probably shouldn't be a type because it's a type flag, and I prefer for it to stand out from the rest of types.
 
> An exception was leaving 'union' as a type instead of a keyword. According to the Hare spec, it seems more like a type than a keyword to me. Let me know if you strongly disagree and I'll change it too.

Both struct and union are types. I'm fine with highlighting them either as types or as keywords, but it should be consistent. The initial version used a keyword for struct, so I went with that for union as well.

orbitalquark:
> I'm a bit skeptical of highlighting regular functions. Can it be configured to only highlight builtins?

You should be able to configure your editor to highlight only builtin functions. By default, Scintillua's SciTE themes do this: https://github.com/orbitalquark/scintillua/blob/ead19010d5390b1ef20bbc176793fc2b435b183a/themes/scite.properties#L45-L47. The Textadept editor also only highlights builtin functions: https://github.com/orbitalquark/textadept/blob/89dfa1ec97ea9ad271cc2366ef5086ab620528e0/themes/light.lua#L51-L53.

> Also, `nullable` probably shouldn't be a type because it's a type flag, and I prefer for it to stand out from the rest of types.

Okay, `nullable` is now a keyword: https://github.com/orbitalquark/scintillua/commit/ead19010d5390b1ef20bbc176793fc2b435b183a. Thanks for the tip.

PR #102, opened 2023-09-14
migrate matlab lexer

rnpnr:


rnpnr:
Sorry I didn't see that builtins (functions and constants) should be split up. I will do that later when I have a chance.

rnpnr:
I split up the functions and variables as recommend in the upgrade legacy
lexers issue. This should be good to go now.

The only problem is that MATLAB is terrible language where variables
are indexed in the same way functions are called. Without language
introspection you can only correctly lex one or the other. So you can
keep it as I have it now and incorrectly lex variables or drop detection
for normal functions and only detect methods and builtin functions.

orbitalquark:
Thanks for this! I've dropped tagging of normal functions (and also methods for the sake of consistency) because tagging variables incorrectly feels wrong. I also added a separate rule for built-in constants like `pi`.

https://github.com/orbitalquark/scintillua/commit/b997a41a9b9f2ed27706e9e79be85db828b5c428 and https://github.com/orbitalquark/scintillua/commit/46f53f4f203d091fe0bff2e8383403abc2692c51

rnpnr:
> I also added a separate rule for built-in constants like `pi`.

Its fine but they are technically variables not constants. You can
reassign values to them. As far as I know there are no constants in
MATLAB.

orbitalquark:
It's the same with Lua, but marking them as constants hopefully encourages good behavior :)

PR #105, opened 2023-10-23
migrate meson lexer

rnpnr:


orbitalquark:
Thank you! I'll take a proper look hopefully this weekend.

lexers/meson.lua
```diff
@@ -1,23 +1,60 @@
...
+-- A method call must be followed by an opening parenthesis.
+lex:add_rule('method', lex:tag(lexer.FUNCTION_METHOD, lex:word_match(lexer.FUNCTION_METHOD) * #(lexer.space^0 * '(')))
+
+-- Function.
+-- https://mesonbuild.com/Reference-manual.html#functions
+-- A function call must be followed by an opening parenthesis. The matching of function calls
+-- instead of just their names is needed to not falsely highlight function names which can also
+-- be keyword arguments. For example 'include_directories' can be a function call itself or a
+-- keyword argument of an 'executable' or 'library' function call.
+lex:add_rule('function', lex:tag(lexer.FUNCTION, lex:word_match(lexer.FUNCTION) * #(lexer.space^0 * '(')))
```
orbitalquark, line 24: These should be `lexer.FUNCTION_BUILTIN`.

lexers/meson.lua
```diff
@@ -1,23 +1,60 @@
...
+
+-- Builtin objects.
+-- https://mesonbuild.com/Reference-manual.html#builtin-objects
+lex:add_rule('object', lex:tag(lexer.VARIABLE_BUILTIN, lex:word_match(lexer.VARIABLE_BUILTIN)))
+
+-- Constants.
+lex:add_rule('constant', lex:tag(lexer.CONSTANT, lex:word_match(lexer.CONSTANT)))
```
orbitalquark, line 31: Similarly, these should be `lexer.CONSTANT_BUILTIN`.

lexers/meson.lua
```diff
@@ -1,23 +1,60 @@
...
+-- Strings.
+local str = lexer.range("'", true)
+local multiline_str = lexer.range("'''")
+lex:add_rule('string', lex:tag(lexer.STRING, multiline_str + str))
+
+-- Comments.
+lex:add_rule('comment', lex:tag(lexer.COMMENT, lexer.to_eol('#', true)))
+
+-- Numbers.
+local oct = '0o' * lexer.oct_num
```
orbitalquark, line 45: This will not match `0o7`. `lexer.oct_num` needs a `0` prefix, so this should be `0o * lpeg.R('07')`.

lexers/meson.lua
```diff
@@ -1,23 +1,60 @@
...
+local integer = S('+-')^-1 * (lexer.bin_num + lexer.dec_num + lexer.hex_num + oct)
```
orbitalquark, line 46: `lexer.hex_num` should be matched first, or else `lexer.dec_num` will eat the `0` in `0xff`, resulting in an unexpected match.

orbitalquark (commented):
Sorry for the delayed review. I will commit it with changes related to the notes I've made in this review, so there's no need for you to do anything. Thanks again!

lexers/meson.lua
rnpnr: Fair, I wasn't sure because Meson only offers builtin functions; a user can't define new functions.

lexers/meson.lua
rnpnr: :+1: for some reason I wasn't thinking about order when I changed those.

orbitalquark:
Updated via https://github.com/orbitalquark/scintillua/commit/48a6fc9511ec67993e43ac7f5a33efc616b7ea32

PR #106, opened 2023-11-04
Move tr literal to regex rule in perl lexer

snoopy:
`y/a/b/` should be highlighted as a regex, same as `tr/a/b/`. Technically, this should already work but it doesn't (at least not for me).
Moving `lit_tr` to the regex rule fixes this for me.


orbitalquark:
Thanks for this. When testing, I found a match error unrelated to this PR caused when 'y' is at the end of input. I'll try and diagnose when I have some time before committing this along with a fix.

orbitalquark:
Committed in https://github.com/orbitalquark/scintillua/commit/6c45f3bc2c3256195282318a8b34549b3041c4c0.

PR #108, opened 2024-03-08
Update Autohotkey lexer

helsmy:
Add fold point and keyword `class` and `extends`.

Detail:
- Fold point: Fold between `{` and `}`. Autohotkey use them to mark code block.
- class related keywords: `class` and `extends`. OOP of Autohotkey.


orbitalquark:
Thanks! Committed as https://github.com/orbitalquark/scintillua/commit/683915b032f0917f0af201e020ef163fdafa090d

PR #110, opened 2024-03-19
Highlight types in number literals for Rust

snoopy:
Some other editors highlight the type part of number literals. E.g. in `456i32` the `i32` part will be highlighted as type instead of a number.
I think this makes a lot of sense and should be the default behavior in Textadept as well.

There is probably a better way to do this. The suggested change does not work with octal number literals (e.g. `0o01234u32`).

What are your thoughts on this?


orbitalquark:
I am in favor of this, and committed https://github.com/orbitalquark/scintillua/commit/b789dde0ff9976f98a18012138c68f165a7e5867.

The octal number highlighting issue is unrelated to this change. The octal pattern does not recognize the '0o' prefix like binary ('0b') and hex ('0x'). That's a separate issue that may be worth looking into.

PR #117, opened 2024-09-10
Add c23 attributes

samuelmarquis:
From: https://github.com/martanne/vis/pull/1207

Minor change to the `ansi_c` parser, adds highlighting support for every [c23 attribute](https://en.cppreference.com/w/c/language/attributes) as well as vendor-specific attributes of the form `[[a::b]]`

### DIFF
--- a/lexers/ansi_c.lua
+++ b/lexers/ansi_c.lua
@@ -52,7 +52,9 @@ local preproc = lex:tag(lexer.PREPROCESSOR, '#' * S('\t ')^0 * lex:word_match(le
 lex:add_rule('preprocessor', include + preproc)
 
 -- Attributes.
-lex:add_rule('attribute', lex:tag(lexer.ATTRIBUTE, '[[' * (lex:word_match(lexer.ATTRIBUTE) + (lexer.word * ':' * lexer.word)) * ']]'))
+lex:add_rule('attribute', lex:tag(lexer.ATTRIBUTE, '[[' * 
+  (lex:word_match(lexer.ATTRIBUTE) + (lexer.word * ':' * lexer.word)) * 
+  lexer.range('(', ')')^-1 * ']]'))
 
 -- Operators.
 lex:add_rule('operator', lex:tag(lexer.OPERATOR, S('+-/*%<>~!=^&|?~:;,.()[]{}')))
--- a/lexers/ansi_c.lua
+++ b/lexers/ansi_c.lua
@@ -53,7 +53,7 @@ lex:add_rule('preprocessor', include + preproc)
 
 -- Attributes.
 lex:add_rule('attribute', lex:tag(lexer.ATTRIBUTE, '[[' * 
-  (lex:word_match(lexer.ATTRIBUTE) + (lexer.word * ':' * lexer.word)) * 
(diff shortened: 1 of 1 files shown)


lexers/ansi_c.lua
```diff
@@ -51,6 +51,11 @@ local include = lex:tag(lexer.PREPROCESSOR, '#' * S('\t ')^0 * 'include') *
 local preproc = lex:tag(lexer.PREPROCESSOR, '#' * S('\t ')^0 * lex:word_match(lexer.PREPROCESSOR))
 lex:add_rule('preprocessor', include + preproc)
 
+-- Attributes.
+lex:add_rule('attribute', lex:tag(lexer.ATTRIBUTE, '[[' * 
+  (lex:word_match(lexer.ATTRIBUTE) + (lexer.word * ':' * lexer.word)) *
```
orbitalquark: I'm pretty sure you meant to use `'::'` instead of `':'`.

lexers/ansi_c.lua
```diff
@@ -51,6 +51,11 @@ local include = lex:tag(lexer.PREPROCESSOR, '#' * S('\t ')^0 * 'include') *
 local preproc = lex:tag(lexer.PREPROCESSOR, '#' * S('\t ')^0 * lex:word_match(lexer.PREPROCESSOR))
 lex:add_rule('preprocessor', include + preproc)
 
+-- Attributes.
+lex:add_rule('attribute', lex:tag(lexer.ATTRIBUTE, '[[' * 
+  (lex:word_match(lexer.ATTRIBUTE) + (lexer.word * ':' * lexer.word)) * 
+  lexer.range('(', ')')^-1 * ']]'))
```
orbitalquark: Do you think there is any value in having annotation arguments highlighted normally? For example, should `[[deprecated("reason")]]` highlight `"reason"` as a string? (And perhaps highlight the parentheses as operators?)

It may be worth trying to highlight arguments separately. If so, we may want to consider highlighting `[[` and `]]` as single entities (either as attributes or operators) and then use `B('[[') * lex:tag(lexer.ATTRIBUTE, lex:word_match(lexer.ATTRIBUTE) + lexer.word * '::' * lexer.word)` to match attributes.

This would not work for attribute lists (e.g. `[[foo, bar, baz]]` would highlight foo as an attribute, but not bar or baz), but it would catch a vast majority of cases.

orbitalquark (changes requested):
Thanks for your contribution! There are a couple things to consider first that I've left feedback for.

lexers/ansi_c.lua
samuelmarquis: so true

lexers/ansi_c.lua
samuelmarquis: I'll defer here. In my use case I'd rather have the entire attribute highlit as one thing. Having the string marked as such makes sense--but imo having the parens marked as well would be too much visual clutter, and doing one without the other would seem arbitrary.

Attribute lists are an important consideration though, I'll add a patch for that momentarily.

orbitalquark:
Fair enough. Thanks for the contribution! Committed in https://github.com/orbitalquark/scintillua/commit/799e8ae29e2cac80a0ff9a039e25a4a758eaf4d1. I found a typo and changed "depreciated" to "deprecated".

PR #17, opened 2020-12-20
Add html lexer performance improvement used by vis

moesasji:
Adds commit baa51e934ce057af5b5be829d6a73a3e8b4c03d0 which is the same performance fix as used by the wsf and xml lexer in vis.

PR #118, opened 2024-10-03
Speedup lexers

mcepl:
This is just rebasing of #15, #16, and #17 on the top of the current master.

I have been running this for the past couple of weeks on the top of vis, and it doesn't seem to break anything. I haven’t measured the speed difference, but even with large XML files (Bible text, approx. 15MB) the editor was very swift.

### DIFF
--- a/lexers/html.lua
+++ b/lexers/html.lua
@@ -28,15 +28,9 @@ lex:add_rule('tag_close', tag_close)
 
 -- Equals.
 -- TODO: performance is terrible on large files.
-local in_tag = P(function(input, index)
-	local before = input:sub(1, index - 1)
-	local s, e = before:find('<[^>]-$'), before:find('>[^<]-$')
-	if s and e then return s > e end
-	if s then return true end
-	return input:find('^[^<]->', index) ~= nil
-end)
+local in_tag = #P((1 - S'><')^0 * '>')
 
-local equals = lex:tag(lexer.OPERATOR, '=') -- * in_tag
+local equals = lex:tag(lexer.OPERATOR, '=') * in_tag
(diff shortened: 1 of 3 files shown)


orbitalquark:

orbitalquark:
Thanks again for submitting this. To be clear, the lexers were already fast, as there was no expensive tag test to highlight '=' and number attribute values -- it could be a bit sloppy.

This PR adds an inexpensive postfix (look-ahead) tag test. The commented out test was a prefix test with look-behind.

After experimentation, there are a couple issues I have with this change.

1. Normally, Scintillua matches entities with the closing delimiter as optional (strings, block comments, etc.). This change goes against that. Something like `<div class="foo"` will not highlight properly until the closing `>` is added. You won't see this if your editor automatically adds a closing '>' when typing '<'.
2. Some editors start lexing at the start of a line, so building on the point above, a multi-line tag will never highlight attributes properly, even after typing a closing '>'. For example:

```
<div
class="foo"
style="float: left;"
```

Typing a '>' on the 'style' line will highlight only that line and leave the 'class' line alone, and typing a '>' after the 'style' line will not highlight either attribute line.

The first issue I could perhaps be okay with, as technically a tag is not valid until it is closed, but the second issue is more problematic, because we cannot assume an editor inserts a trailing '>' automatically, so Scintillua may not highlight a finished tag.

mcepl:
I have no problem if you decide that you don’t want these changes (I am quite certain, that you understand intricacies of Scintillua lexers a way more than me and @moesasji combined), but then it would be probably better to make the decision and close all those PRs.

orbitalquark:
Okay, I will close the PRs then.

PR #119, opened 2024-10-04
Adding Factor lexer

mrjbq7:
The [Factor programming language](https://github.com/factor/factor) project has had a copy of a lexer for Factor that should probably be upstreamed.

https://github.com/factor/factor/blob/881aa979b86034a9330db9f372e7c6cd10a313f8/misc/textadept/lexers/factor.lua

I pulled it out for upstreaming.

### DIFF
--- a/lexers/factor.lua
+++ b/lexers/factor.lua
@@ -1,0 +1,104 @@
+--[[
+Lexer for the Factor programming language (http://factorcode.org)
+Copyright 2013 Michael T. Richter <ttmrichter@gmail.com>
+
+This program is free software and comes without any warranty, express nor
+implied.  It is, in short, warranted to do absolutely nothing but (possibly)
+occupy storage space.  You can redistribute it and/or modify it under the terms
+of the Do What The Fuck You Want To Public License, Version 2, as published by
+Sam Hocevar.  Consult http://www.wtfpl.net/txt/copying for full legal details.
+
+BUGS
+====
+At this time the lexer is usable, but not perfect.  Problems include:
+ * identifiers like (foo) get treated and coloured like stack declarations
+ * other as-yet unknown display bugs  :-)
+
+These make a few source files less than lovely and will be fixed as possible.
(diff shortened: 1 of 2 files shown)


orbitalquark:
Thanks for your contribution! This is a lexer written in a really old format that will need to be migrated before it can be added. There will be two-steps in this process. First:

https://github.com/orbitalquark/scintillua/blob/64f3a6a42493139c836238a6df2837739f739c48/lexers/lexer.lua#L460-L571

Then: https://orbitalquark.github.io/scintillua/api.html#migrating-legacy-lexers

I can do this for you, but it will be in my own time.

mrjbq7:
Okay great, I will do this in a week or so.

mrjbq7:
Okay, I updated the lexer from legacy, and added a test.

Perhaps a second pass would improve a few things about syntax highlighting for [Factor](https://factorcode.org) but it's quite usable as-is.

Some comments:

1) tried to use ``make deps`` but it had an error checking the certificate for ``lpeg-1.1.0.tar.gz``, so I briefly tried this:

```diff
-$(lpeg_tgz): ; $(WGET) http://www.inf.puc-rio.br/~roberto/lpeg/$@
+$(lpeg_tgz): ; $(WGET) --no-check-certificate http://www.inf.puc-rio.br/~roberto/lpeg/$@
```

2) but it didn't seem to build the deps properly, so I just switched to using ``luarocks install lpeg`` and ``luarocks install luafilesystem``.

3) ``make tests`` only has a single failure in ``test_lua51`` which is only because I don't have it installed on my system.

Is there anything else you need done to this?

mrjbq7:
And I just noticed the "second pass" request, which includes ``lexer.new(...)``, anyway I did that also.

orbitalquark:
Thank you! I will review and add this when I have some time.

orbitalquark:
Added in https://github.com/orbitalquark/scintillua/commit/043379192a3a96abe6968b14317c03c8bd00bd54

PR #122, opened 2024-12-21
Add Nix lexer

### LINKED ISSUE #75: Nix lexer
Would be great to have Nix lexer. As NixOS and vis user I really miss it.
See https://github.com/martanne/vis/issues/880

samuelmarquis:
Fixes #75

Turns out, there's no formal specification for the Nix language, which made this not at all frustrating, so I referenced the Nix [tree-sitter](https://github.com/nix-community/tree-sitter) (or what I could make of it) and the [official lexer](https://github.com/NixOS/nix/blob/7a71621b7c43d7d2f264cc495fb7ceb66455fd3c/src/libexpr/lexer.l) in constructing this.

I wanted paths tagged, but didn't see a super close fit in the list of standard tags. Paths are kind of like URIs, so, they all get grouped in links? Why not. I also broke out a few of the operators--`&&` is fine, `&` isn't.

There's a _ton_ of builtin functions/constants, but they're mostly buried as members inside a global `builtins` object (so you have to do something like `builtins.sort`), so I only included the ones you can directly refer to.

The syntax highlighting in vim grabs the last word in a chain of member accesses, i.e. boot.loader.systemd-boot.**enable**, but I elected not to do this, as it doesn't feel so crucial.

There's also the matter of `"strings ${with command substitutions}}"`. This seemed gross to deal with (and the bash parser doesn't bother) so I've left it out, the whole thing is a string.

I'm admittedly not 100% sure why things like `;` and brackets are considered operators, but that's the way everything else seems to do it.

I also threw the necessary changes into `scintillua.properties` and `lexer.lua`.


lexers/nix.lua
```diff
@@ -0,0 +1,81 @@
...
+-- Comments.
+local line_comment = lexer.to_eol('#', true)
+local block_comment = lexer.range('/*', '*/')
+lex:add_rule('comment', lex:tag(lexer.COMMENT, line_comment + block_comment))
+
+-- Numbers.
+lex:add_rule('number', lex:tag(lexer.NUMBER, lexer.number))
+
+-- Operators.
+local ops = S('?+-.*/!<>=,;:()[]{}') + lex:word_match(lexer.OPERATOR)
```
orbitalquark: Please use `P('&&') + '||' + ...` instead of word match because these characters will be added to an internal word character list and result in more cases of potential word matches and then backtracking on failure.

Also, if you really want to match '//', '->', and '++' on their own, they need to be matched before the `S()` or else '//' will be lexed as two separate '/' characters.

lexers/nix.lua
```diff
@@ -0,0 +1,81 @@
...
+	'let', 'in', 'rec', 'inherit', 'or', '...',
+})
+
+lex:set_word_list(lexer.CONSTANT_BUILTIN, {
+	'builtins', 'true', 'false', 'null',
+})
+
+-- Directly accessible functions.
+lex:set_word_list(lexer.FUNCTION_BUILTIN, {
+	'derivation', 'import', 'abort', 'throw',
```
orbitalquark, line 99: When you say there are a ton of Nix builtin functions and contants, how many are we talking? I can see value in having them if they are part of the language.

For example, the C++ lexer includes over 100 stl functions and types: https://github.com/orbitalquark/scintillua/blob/default/lexers/cpp.lua. It may not be exhaustive, but it's quite helpful for editing a wide range of C++ code.

orbitalquark (changes requested):
This looks really good, thank you! I've left a couple of comments for you to address when you get the chance.

lexers/nix.lua
samuelmarquis: So does changing the order of addition here provide this effect (i.e. `P('&&') + P('||') + ... P('->') + S('?+...{}')` (does only the first need to be in a `P()` call?)), or do I want a separate, prior `add_rule()` call for the long operators?

lexers/nix.lua
samuelmarquis: The list of constants is [here](https://nix.dev/manual/nix/2.18/language/builtin-constants) and functions [here](https://nix.dev/manual/nix/2.18/language/builtins); and there are many more that are frequently used which live inside the almost-always included `pkgs` object and its members. I'll see if I can find a way to automate dropping all of them into a list, because the prospect of hand-copying them seems quite tedious--but I agree it'd be valuable to have them.

lexers/nix.lua
orbitalquark: Yes, switching the operands to `+` is sufficient. It's an ordered choice operator, so it tries to match left first. If that fails, it tries to match right.

lexers/nix.lua
orbitalquark: Yeah, I'd like to have them if it's not too much trouble. Hopefully there's an automated way to do it. Thanks!

lexers/nix.lua
samuelmarquis: Is it left-associative? Wrt `P() + "..." + S()`

orbitalquark: Yes, I think so.

lexers/nix.lua
samuelmarquis: Got all the builtins, probably going to come back to this in the future and add everything in `pkgs` and `pkgs.libs`/`pkgs.stdenv`, but those docs pages were a lot less trivial to scrape

lexers/nix.lua
samuelmarquis: Solid, refactored this

lexers/nix.lua
orbitalquark: Okay, however now the lexer will highlight bare `add()` as a function, when it should be `builtins.add()`, right?

You'll have to modify your `function` rule to use two word lists, the first one is for non-builtins, and the second one is for builtins. For example, `lex:add_rule('function', lex:tag(lexer.FUNCTION_BUILTIN, lex:word_match(lexer.FUNCTION_BUILTIN) + 'builtins.' * lex:word_match(lexer.FUNCTION_BUILTIN .. '.builtins'))` where `lex:set_word_list(lexer.FUNCTION_BUILTIN .. '.builtins', { ... }` contains `'add'`, etc.

If you're tired of working on this I can take over when I have some time. I think I have enough context to push this over the finish line.

lexers/nix.lua
```diff
@@ -0,0 +1,101 @@
...
+local lexer = lexer
+local P, S = lpeg.P, lpeg.S
+
+local lex = lexer.new(...)
+
+-- Keywords.
+lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD)))
+
+-- Functions.
+lex:add_rule('function', lex:tag(lexer.FUNCTION_BUILTIN, lex:word_match(lexer.FUNCTION_BUILTIN)))
```
orbitalquark: I just thought of this now, but would it make sense to only highlight functions if a '(' follows? This is the case for many lexers.

orbitalquark (changes requested):
Almost there!

lexers/nix.lua
samuelmarquis: Gonna have to double-check and make sure there's no Haskell-style calling convention weirdness, but that's probably not a bad idea

lexers/nix.lua
samuelmarquis: No problem, I'll add this in a moment

samuelmarquis: It does seem mildly sub-ideal to consume the `.` in `builtins.` so I'll see about matching without consuming; that way `builtins` can still match as a constant in its own right when it's used to get a function.

orbitalquark: On the contrary, I would like it to consume `builtins.` as part of the function, just like the Lua lexer does with things like `string.find()`.

samuelmarquis: So would, for example, `builtins.currentSystem` consume the entire thing as a constant? This feels slightly dubious to me wrt `builtins` getting parsed two different ways and losing the `.` operator, but I'll defer if this desired.

orbitalquark: Yes, `builtins.currentSystem` would be all highlighted as a constant, including the `.` operator. The primary reason is for backtracking. In case clients want to start syntax highlighting from the current position (e.g. for performance), backtracking to a style change is sufficient. If we broke up `builtins.currentSystem`, then a client would backtrack to just the `.`, would start highlighting, but would not see `currentSystem` as belonging to `builtins`, and would highlight incorrectly. By highlighting it all, any backtracking would allow highlighting to start at `builtins`, catching any changes at the current position. I hope this makes sense.

Does Nix allow using `builtins` by itself? (e.g. passing as a parameter) If it doesn't, then I probably wouldn't highlight `builtins` by itself since it may be meaningless and give that hint to the user.

samuelmarquis: I have yet to see it in the wild... but every day I encounter a new strange usage of this language that is foreign to me. I'll do a little digging but we can omit it for now, I'll make this and the other change

lexers/nix.lua
samuelmarquis: [Yeah doesn't look like this works, it's Haskell-style.](https://nix.dev/manual/nix/2.18/language/builtins)

samuelmarquis:
See how you feel about the `pkgs` dilemma & the identifier rule, but I'm pretty indifferent whichever way you want to go in resolving those.

lexers/nix.lua
samuelmarquis: Didn't find anything doing this in my cursory search, but it's nice to have highlit while typing I guess? If I later find out it's illegal I can make another PR to remove it, but I'm honestly ambivalent on that detail.

samuelmarquis:
My basic thought with the `pkgs` thing is: it'd be nice for lists of packages like this to highlight every package_name `(alnum + '_-')` inside the brackets (sans the brackets themselves, comments, etc)
image
but NOT grab any other Nix expression, so immediate escape on parens, a function name, whatever.
image
^this is bad

 _Some_ but maybe not _all_ operators are fine here. Is there a non-trivial way to do this? I'm guessing no, so feel free to remove that element and I might come back to it later if I have a better idea of how to get the effect I want.

lexers/nix.lua
samuelmarquis: One more thought about this--people pretty frequently will import specific names they plan on using from `builtins` to their environment, and imo those should still get lexed, so I'll throw a `^0` on the `P('builtin.')`

samuelmarquis: Just now realizing this might make the backtracking thing moot, as--to my understanding of what you've written at least--a highlight of `currentSystem` _would_ still be correct given how often that name & others like it are imported.

Does seem like the operator `.` in vis still highlights separately (which makes sense now that I'm looking at it) so I'm happy either way wrt that detail

orbitalquark:
Thanks for all your hard work here!!! This PR is getting pretty messy (not your fault), so I cleaned it up a bit and committed an initial version here: https://github.com/orbitalquark/scintillua/commit/a864eb0a242e6d3716df3e370e1d860a17576375.

We can iterate on it in a new PR as you see fit.

Here's what I did:

- I went ahead and moved all of the builtin functions and constants into their respective wordlists, _including_ a `builtins.` prefix as necessary. Yes, it's repetitive, but honestly it feels cleaner than maintaining two word lists. You mentioned that some people import specific names and end up using bare names without the `builtin.` prefix. If this is indeed common, then we may want to address it, but I'm hesitant about making the `builtin.` optional, as bare words will match, even if they are used as just variable names. That is sub-optimal.
- I left out the `lexer.range('[', ']')` bit because that will match lists like `x = [1 2 3]`, which I think is pretty undesirable. My first thought is that what you want to do may not be possible with Scintillua in its current state. I could be wrong though. We may want to explore this further.
- I reverted your identifier rule to your first iteration because `id_char^1` will match numbers before your numbers rule.

PR #123, opened 2024-12-21
Migrate Haskell lexer

samuelmarquis:
Migrate the Haskell lexer to the new style.


orbitalquark:
Thank you! Committed in https://github.com/orbitalquark/scintillua/commit/afdd70877adacde04b36a2d1768c26dc86d72aff.

PR #125, opened 2024-12-22
C adjustments

samuelmarquis:
The `ssize_t` stuff in the first commit needs no real explanation.

However; the Linux kernel stuff is ... more dubious.

It would be _nice_ to have a comprehensive list of types, constants, functions, and macros that are in the kernel's various headers; unsurprisingly, there are thousands of them, so including them _all_ is infeasible; I'm frightened to think of what it would do to performance. This commit adds only the types in `include/linux/types.h`, and the corresponding constants in `include/linux/limits.h`.

At some point down the road, I may take a closer look through the source and send another pull request once I'm better able to determine what the common vs uncommon members are, and create a narrower selection of the important ones. `kmalloc()`? Good to have. `phy_interface_set_rgmii()`? Don't think anyone will miss it.

Speaking of things nobody will miss--I'm going to do a little more digging and then see if `U32_MIN` can really be removed. May not even be worth bothering to include it when you make this commit--there are zero references to it anywhere in the kernel.

I normally use vis to test these adjustments; but I'm not sure how to reconcile the "two groups of thing in the same tag" problem, as it disabled syntax highlighting for both groups--maybe that's on their side & they should deal with it in their upcoming merge, maybe it's here. Let me know how you want that handled.

### DIFF
--- a/lexers/c.lua
+++ b/lexers/c.lua
@@ -43,7 +43,7 @@ lex:add_rule('comment', lex:tag(lexer.COMMENT, line_comment + block_comment))
 
 -- Numbers.
 local integer = lexer.integer * lexer.word_match('u l ll ul ull lu llu', true)^-1
-local float = lexer.float * P('f')^-1
+local float = lexer.float * lexer.word_match('f l df dd dl i j', true)^-1
 lex:add_rule('number', lex:tag(lexer.NUMBER, float + integer))
 
 -- Preprocessor.


samuelmarquis:
Added another commit to include some missing floating point literal suffixes (suffi?).

lexers/c.lua
```diff
@@ -11,6 +11,7 @@ lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD)))
 
 -- Types.
 lex:add_rule('type', lex:tag(lexer.TYPE, lex:word_match(lexer.TYPE)))
+lex:add_rule('linux-types', lex:tag(lexer.TYPE .. '.linux', lex:word_match(lexer.TYPE .. '.linux')))
```
orbitalquark: I would prefer not to have domain-specific things in a standard C lexer. See comment here: https://github.com/orbitalquark/scintillua/discussions/124#discussioncomment-11643726

Sorry about that.

lexers/c.lua
```diff
@@ -22,7 +23,8 @@ lex:add_rule('function', (builtin_func + method + func) * #(lexer.space^0 * '(')
 -- Constants.
 lex:add_rule('constants', lex:tag(lexer.CONSTANT_BUILTIN,
 	-(B('.') + B('->')) * lex:word_match(lexer.CONSTANT_BUILTIN)))
-
+lex:add_rule('linux-constants', lex:tag(lexer.CONSTANT_BUILTIN .. '.linux',
```
orbitalquark: I would prefer not to have domain-specific things in a standard C lexer. See comment here: https://github.com/orbitalquark/scintillua/discussions/124#discussioncomment-11643726

Sorry about that.

lexers/c.lua
```diff
@@ -43,7 +45,9 @@ lex:add_rule('comment', lex:tag(lexer.COMMENT, line_comment + block_comment))
 
 -- Numbers.
 local integer = lexer.integer * lexer.word_match('u l ll ul ull lu llu', true)^-1
-local float = lexer.float * P('f')^-1
+local float = lexer.float * (P('f')+'l'+'F'+'L'+'df'+'dd'+'dl'+'DF'+'DD'+'DL'
```
orbitalquark: What do you think about using a `lexer.word_match()` like for integers above? It will allow for case-insensitive matching, which ought to be good enough. (It's a lexer, not a compiler.)

lexers/c.lua
```diff
@@ -208,6 +214,31 @@ lex:set_word_list(lexer.ATTRIBUTE, {
...
+	'PHYS_ADDR_MAX', 'RESOURCE_SIZE_MAX', 'U8_MAX', 'S8_MAX', 'S8_MIN',
+	-- There is a #define U32_MIN ((u32)o), but not any of the others.
+	-- I'll refrain from going on a tirade here, but suffice to say it exists
+	-- to do codegen, but there don't appear to be any live references to it.
+	-- I'm going to see if it's really useless, and send a patch to remove
+	-- it if so. Feel free to not include it, as it may be gone fairly soon.
+	'U16_MAX', 'S16_MAX', 'S16_MIN', 'U32_MAX', 'U32_MIN',
+	'S32_MAX', 'S32_MIN', 'U64_MAX', 'S64_MAX', 'S64_min',
+})
+
```
orbitalquark: I would prefer not to have domain-specific things in a standard C lexer. See comment here: https://github.com/orbitalquark/scintillua/discussions/124#discussioncomment-11643726

Sorry about that.

orbitalquark (changes requested):
Other than the domain-specific stuff I mentioned in the discussion thread, it looks like there's just one minor thing to address here. After that I think this should be okay.

lexers/c.lua
samuelmarquis: Yeah, that'd work, I'll make the change

samuelmarquis:
Nuked the non-portable stuff, brought the FP literal suffi in line with the ints.

orbitalquark:
Thanks very much. Committed in https://github.com/orbitalquark/scintillua/commit/14d10e430e13e143bbcbd122ab9fa4616b11d4d7.

Sorry for all the trouble.

PR #126, opened 2024-12-27
Fix spelling of my name

mcepl:
I understand that normal humans cannot comprehend that Slavic mishmash we made of the Latin alphabet, but this is my real name.


orbitalquark:
Yes, of course. Fixed in https://github.com/orbitalquark/scintillua/commit/48a5ff337b05849f2bc21ba31f9961bee7cab190. Sorry about that!

PR #127, opened 2025-01-15
Added lexer for org-mode format (originally for Emacs)

mcepl:
The webpage of that project is https://orgmode.org; this lexer is almost complete rewrite of the original from https://github.com/rgieseke/textadept-orgmode:

1. separated into issue https://github.com/orbitalquark/scintillua/issues/128

2. Is there some way how to modify this lexer from the user configuration? I have there now that SUSE abbreviations pattern to include into links, and I can see there are others (https://github.com/martanne/vis/issues/1228, although this might be better to be changed inside of this repo lexers).

### DIFF
--- a/lexers/org.lua
+++ b/lexers/org.lua
@@ -1,0 +1,101 @@
+--[[ Textadept Org module -- Org agenda LPeg lexer
+Copyright (c) 2012 joten
+Copyright (c) 2025 Matěj Cepl (@mcepl everywhere)
+Released under the conditions described in the LICENSE file.
+--]]
+
+local lexer = lexer
+local word_match = lexer.word_match
+local P, R, S, V = lpeg.P, lpeg.R, lpeg.S, lpeg.V
+
+local lex = lexer.new(...)
+
+--[[Overview. Used examples.
+* Heading 1-5 (8)
+	TODO special color, bold
+	DONE special color, bold
+	[#priority]
+	:tag: bold
+<date> / [date] like heading 3, underlined
+	keywords: CLOSED: DEADLINE: SCHEDULED: like heading 3, not underlined
+| table | like heading 1
+[[link][description] ]
+formatting:
+	*bold*
+	/italic/
+	_underline_
+	+strike+
+--]]
+
+-- Font formats.
+lex:add_rule('bold', lex:tag('BOLD',      '*' * lexer.word * '*'))
+lex:add_rule('italic', lex:tag('ITALIC',    '/' * lexer.word * '/'))
+lex:add_rule('underline', lex:tag('UNDERLINE', '_' * lexer.alnum * '_'))
+
+-- ToDos.
+lex:add_rule('todo', lex:tag('TODO', lex:word_match('TODO')))
+lex:add_rule('done', lex:tag('DONE', lex:word_match('DONE')))
+lex:add_rule('wontfix', lex:tag('WONTFIX', lex:word_match('WONTFIX')))
+
+-- DateTime.
+local DD =   lexer.digit * lexer.digit
+local time_range = (' ' * DD * ':' * DD)^0 * ('-' * DD * ':' * DD)^0
+local repeater =   (' +' * lexer.integer * S('dwmy'))^0
+local alarm =      (' -' * lexer.integer * S('dwmy'))^0
+local pattern_datetime1 = S('<[')^-1 * 'date' * 'wday' * time_range * repeater * alarm * S('>]')^-1
+local pattern_datetime2 = lexer.starts_line('weekday') * lexer.space^1 * DD * '. ' * 'month' * lexer.space^1 * DD * DD
+local datetime =   pattern_datetime1 + pattern_datetime2
+
+lex:add_rule('current_date', lex:tag('CURRENT_DATE', lexer.starts_line(lex:word_match('weekday')) * lexer.space^1 * DD * '. ' * lex:word_match('month') * lexer.space^1 * DD * DD * '|'))
+lex:add_rule('time', lex:tag(lexer.CLASS, DD * ':' * DD))
+lex:add_rule('week', lex:tag('UNDERLINE', lexer.starts_line('KW ' * DD * lexer.space^25) + lexer.starts_line('Wk ' * DD * lexer.space^25)))
+lex:add_rule('datetime', lex:tag(lexer.NUMBER, lex:word_match('datetime')))
+
+-- Heading patterns.
+local function h(n)
+	return lex:tag(string.format('%s.h%s', lexer.HEADING, n),
+		lexer.starts_line(P(string.rep('*', n)) * ' '))
+end
+lex:add_rule('header', h(6) + h(5) + h(4) + h(3) + h(2) + h(1))
+
+-- Links.
+local orgmode_link = '[[' * (lexer.nonnewline - ' ' - ']')^1 * ']' * ('[' * (lexer.nonnewline - ']')^1 * ']')^0 * ']'
+lex:add_rule('link', lex:tag(lexer.LINK, orgmode_link))
+
+-- Strings.
+lex:add_rule('string', lex:tag(lexer.STRING, P('L')^-1 * lexer.range('"')))
+
+-- Comments.
+local line_comment = lexer.starts_line(lexer.to_eol('# '))
+local block_comment = lexer.range(lexer.starts_line('#+BEGIN_COMMENT'),
+		lexer.starts_line('#+END_COMMENT'))
+lex:add_rule('comment', lex:tag(lexer.COMMENT, block_comment + line_comment))
+
+-- Word lists.
+lex:set_word_list('TODO', {'TODO', 'DELEGATED', 'WAITING'})
+
+lex:set_word_list('DONE', {'DONE'})
+
+lex:set_word_list('WONTFIX', {'WONTFIX', 'INVALID'})
+
+lex:set_word_list('wday', {
+	'Mo', 'Di', 'Mi', 'Do', 'Fr', 'Sa', 'So',
+	'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun',
+	'Po', 'Út', 'St', 'Čt', 'Pá', 'So', 'Ne'
+})
+
+lex:set_word_list('weekday', {
+	'Montag', 'Dienstag', 'Mittwoch', 'Donnerstag', 'Freitag', 'Samstag', 'Sonntag',
+	'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday', 'Sunday',
+	'Pondělí', 'Úterý', 'Středa', 'Čtvrtek', 'Pátek', 'Sobota', 'Neděle'
+})
+
+lex:set_word_list('month', {
+	'Januar', 'Februar', 'März', 'April', 'Mai', 'Juni', 'Juli', 'August', 'September', 'Oktober', 'November', 'Dezember',
+	'January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November', 'December',
+	'Leden', 'Únor', 'Březen', 'Duben', 'Květen', 'Červen', 'Červenec', 'Srpen', 'Září', 'Říjen', 'Listopad', 'Prosince'
+})
+
+lexer.property['scintillua.comment'] = '#'
+
+return lex
--- a/docs/lexerlist.md
+++ b/docs/lexerlist.md
@@ -96,6 +96,7 @@ used if [`fold.scintillua.by.indentation`][] is enabled. For SciTE, see *lexers/
 1. Objeck`*`
 1. Objective-C`*`
 1. OCaml
+1. Org
 1. Pascal
 1. Perl`*`
 1. PHP`*`
--- a/lexers/lexer.lua
+++ b/lexers/lexer.lua
@@ -1684,6 +1684,7 @@ function M.detect(filename, line)
 		nsh = 'nsis', nsi = 'nsis', nsis = 'nsis', --
 		obs = 'objeck', --
 		m = 'objective_c', mm = 'objective_c', objc = 'objective_c', --
+		org = 'org', --
 		caml = 'caml', ml = 'caml', mli = 'caml', mll = 'caml', mly = 'caml', --
 		dpk = 'pascal', dpr = 'pascal', p = 'pascal', pas = 'pascal', --
 		al = 'perl', perl = 'perl', pl = 'perl', pm = 'perl', pod = 'perl', --


orbitalquark:
1. I am okay with simply omitting single quoted string delimiters if you think this is a good idea and covers a majority of use cases.
2. If I'm understanding you correctly, users may just want to add their own keywords or words to highlight. If that's the case, I would probably just add a new word list (that is not defined/set). Clients can add to it or set it on demand after loading the lexer.

mcepl:
>     2. If I'm understanding you correctly, users may just want to add their own keywords or words to highlight. If that's the case, I would probably just add a new word list (that is not defined/set). Clients can add to it or set it on demand after loading the lexer.

I was talking about 

```lua
local suse_abbrev = ( lpeg.lower^-4 * S("#-") )^1 * lpeg.digit^1
...
```

I have removed `suse_abbrev` from my `org.lua` and added to my global vis configuration script this:

```lua
if win.syntax == 'org' then
	local lex = vis.lexers.load(win.syntax)
	local nonnewline = vis.lexers.nonnewline
	local suse_abbrev = lpeg.B(lpeg.space) * ( lpeg.lower^-4 * S("#-") )^1 * lpeg.digit^1 * #lpeg.space
	local orgmode_link = lpeg.P('[[') * (nonnewline - ' ' - ']')^1 * ']' * ('[' * (nonnewline - ']')^1 * ']')^0 * ']'
	lex:modify_rule('link', lex:tag(vis.lexers.FUNCTION, orgmode_link + suse_abbrev))
end
```

It is syntactically correct, the code is loaded with `org` documents (when I omitted `lpeg.` for `S`, the editor failed), but SUSE abbreviations are not highlighted (and they were when the code was in `org.lua` as in this PR now). Any idea, what could be wrong (of course, you are not responsible for possible bugs in vis)?

Cc: @rnpnr, @fischerling, @jeremybobbin, @Shugyousha … hi, guys, do you see what I am doing wrong with this code in my `visrc.lua`, please?

Just to be sure, `lexer.lua` is almost identical:

```diff
--- /home/matej/build/mcepl/vis/vis/lua/lexers/lexer.lua	2025-01-08 15:39:27.028067594 +0100
+++ /home/matej/repos/scintillua/lexers/lexer.lua	2025-01-15 17:41:13.262440877 +0100
@@ -1,4 +1,4 @@
--- Copyright 2006-2024 Mitchell. See LICENSE.
+-- Copyright 2006-2025 Mitchell. See LICENSE.
 
 --- Lexes Scintilla documents and source code with Lua and LPeg.
 --
@@ -1444,7 +1444,7 @@
 --- Creates a substitute for some Scintilla tables, functions, and fields that Scintillua
 -- depends on when using it as a standalone module.
 local function initialize_standalone_library()
-	M.property = setmetatable({['scintillua.lexers'] = package.path:gsub('/%?%.lua', '/lexers')}, {
+	M.property = setmetatable({['scintillua.lexers'] = package.path:gsub('/%?%.lua', '')}, {
 		__index = function() return '' end, __newindex = function(t, k, v) rawset(t, k, tostring(v)) end
 	})
```

orbitalquark:
I think your `lex:modify_rule()` is fine. I cannot think of a way to extend it otherwise.

mcepl:
Well, fine it might be, but it doesn’t work :crying_cat_face: .

orbitalquark:
You are creating and modifying `lex` local to your `if` conditional, but you don't appear to be doing anything with it. I don't know how vis works though. Does `vis.lexers.load()` return a cached version? If so, after you modify a lexer, make sure you're re-lexing.

mcepl:
@rnpnr are you able to help here? How to re-lex current document? `window:draw()`? `window:set_syntax(window.syntax)`?

Or, @orbitalquark , is it somehow possible to rerun `lex:lex()` on the whole current document? What exactly should be value of `text` and `init_style` in such case?

orbitalquark:
The client application is responsible for invoking the lexer's `lex()` method and set styles based on the returned table. The lexer cannot do this by itself.

fischerling:
> @rnpnr are you able to help here? How to re-lex current document? `window:draw()`? `window:set_syntax(window.syntax)`?

`vis:win:draw()` does redraw the current window and thus relexes the viewport.
This is what `vis-spellcheck` does when invoking one of its commands.

mcepl:
> `vis:win:draw()` does redraw the current window and thus relexes the viewport. This is what `vis-spellcheck` does when invoking one of its commands.

Does it? Any idea what’s wrong here then?

```lua
vis.events.subscribe(vis.events.WIN_OPEN, function(win)
	-- Your per window configuration options e.g.
	if win.syntax == 'org' then
		vis:info("Modifying org lexer")
		local lex = vis.lexers.load(win.syntax)
		local nonnewline = vis.lexers.nonnewline
		local suse_abbrev = lpeg.B(lpeg.space) * ( lpeg.lower^-4 * lpeg.S("#-") )^1 * lpeg.digit^1 * #lpeg.space
		local orgmode_link = lpeg.P('[[') * (nonnewline - ' ' - ']')^1 * ']' * ('[' * (nonnewline - ']')^1 * ']')^0 * ']'
		lex:modify_rule('link', lex:tag(vis.lexers.FUNCTION, orgmode_link + suse_abbrev))
		win:draw()
	end
end)
```

With `lexers/org.lua` from this PR, vis from `master`, this code in `~/.config/vis/visrc.lua`, org file from https://mcepl.fedorapeople.org/tmp/TODO.org, and all those SUSE abbreviations (e.g., `bsc#1235076` on line 10) are not highlighted. Info message is shown correctly, so the code should be activated.

However, that’s just `vis` problem, which will be resolved somehow. This PR is ready for review.

lexers/org.lua
```diff
@@ -0,0 +1,137 @@
...
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program. If not, see <http://www.gnu.org/licenses/>.
```
orbitalquark: Are you okay signing this lexer over to Scintillua's MIT license? I don't want to have different lexers under difference licenses.

lexers/org.lua
```diff
@@ -0,0 +1,137 @@
...
+--]]
+
+local lexer = lexer
+local word_match = lexer.word_match
+local P, R, S, V = lpeg.P, lpeg.R, lpeg.S, lpeg.V
+lpeg.locale(lpeg)
```
orbitalquark: Please remove this line, replace `lpeg.space` with `lexer.space`, and double-check that your patterns still work as expected.

lexers/org.lua
```diff
@@ -0,0 +1,137 @@
...
+formatting:
+	*bold*
+	/italic/
+	_underline_
+	+strike+
+--]]
+
+-- Font formats.
+lex:add_rule('bold', lex:tag('BOLD',      '*' * lexer.word^1 * '*'))
+lex:add_rule('italic', lex:tag('ITALIC',    '/' * lexer.word^1 * '/'))
```
orbitalquark: Is whitespace allowed inside bold and italic delimiters? `lexer.word^1` looks wrong and I think is functionally identical to just `lexer.word` by itself in these contexts.

lexers/org.lua
```diff
@@ -0,0 +1,137 @@
...
+lex:add_rule('underline', lex:tag('UNDERLINE', '_' * lexer.alnum^1 * '_'))
+
+-- ToDos.
+lex:add_rule('todo', lex:tag('TODO', lex:word_match('TODO')))
+lex:set_word_list('TODO', {'TODO', 'DELEGATED', 'WAITING'})
```
orbitalquark: Please move these and other `lex:set_word_list()` calls to the bottom of the lexer in their own section like the other lexers.

lexers/org.lua
```diff
@@ -0,0 +1,137 @@
...
+
+-- XXX I don’t care about priority, if anybody cares, send me a patch
+-- Also, I don’t think that the TAG is the right name, just use a custom tag
+-- local priority = lex:tag('PRIORITY', '[#' * S('ABC') * ']')
+-- local tags = lex:tag('TAG', ':' * (lexer.word + lexer.punct)^1 * lexer.newline)
+
+-- XXX not sure what it is good for
+-- Keywords.
+-- lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD)))
+-- lex:set_word_list(lexer.KEYWORD, {'SCHEDULED', 'DEADLINE', 'CLOSED'})
```
orbitalquark: This unused code should probably be removed to clean things up.

lexers/org.lua
```diff
@@ -0,0 +1,137 @@
...
+local date = DD * DD * '-' * DD * '-' * DD
+lex:set_word_list('wday', {
+	'Mo', 'Di', 'Mi', 'Do', 'Fr', 'Sa', 'So',
+	'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun',
+	'Po', 'Út', 'St', 'Čt', 'Pá', 'So', 'Ne'
+})
+lex:set_word_list('weekday', {
+	'Montag', 'Dienstag', 'Mittwoch', 'Donnerstag', 'Freitag', 'Samstag', 'Sonntag',
+	'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday', 'Sunday',
+	'Pondělí', 'Úterý', 'Středa', 'Čtvrtek', 'Pátek', 'Sobota', 'Neděle'
```
orbitalquark: I never thought to use word lists for localization, but this is a great idea :)

lexers/org.lua
```diff
@@ -0,0 +1,137 @@
...
+local pattern_priority = '[#' * S('ABC') * ']'
+local pattern_tags = ':' * (lexer.word + lexer.punct)^1 * lexer.newline
+local pattern_h1 = lexer.starts_line('* ')        -- Heading1.
+local pattern_h2 = lexer.starts_line('** ')       -- Heading2.
+local pattern_h3 = lexer.starts_line('*** ')      -- Heading3.
+local pattern_h4 = lexer.starts_line('**** ')     -- Heading4.
+local pattern_h5 = lexer.starts_line('***** ')    -- Heading5.
+
+-- Heading lex:tag parts.
+local part_h1 = lex:tag(lexer.CLASS,        (lexer.nonnewline - S('<[') - (':' * lexer.word))^0) * datetime^0 * lex:tag(lexer.CLASS,        (lexer.nonnewline - (':' * lexer.word))^0)
```
orbitalquark: For this and all header-related tags, would you mind using the `lexer.HEADING` tag? For example, Markdown's h1 is `lexer.HEADING .. '.h1'`, h2 is `lexer.HEADING .. '.h2'`, etc.

Tags should be descriptive, not based on other tags that have the desired style/color in the current theme.

lexers/org.lua
```diff
@@ -0,0 +1,137 @@
...
+local part_h5 = lex:tag(lexer.CONSTANT,     (lexer.nonnewline - S('<[') - (':' * lexer.word))^0) * datetime^0 * lex:tag(lexer.CONSTANT,     (lexer.nonnewline - (':' * lexer.word))^0)
+-- Headings.
+local h1 = lex:tag(lexer.CLASS, pattern_h1) *        (agenda_tags * lpeg.space)^0 * (lex:tag('PRIORITY1', pattern_priority) * lpeg.space)^0 * part_h1 * (lex:tag('TAG1', pattern_tags))^0
+local h2 = lex:tag(lexer.STRING, pattern_h2) *       (agenda_tags * lpeg.space)^0 * (lex:tag('PRIORITY2', pattern_priority) * lpeg.space)^0 * part_h2 * (lex:tag('TAG2', pattern_tags))^0
+local h3 = lex:tag(lexer.FUNCTION, pattern_h3) *     (agenda_tags * lpeg.space)^0 * (lex:tag('PRIORITY3', pattern_priority) * lpeg.space)^0 * part_h3 * (lex:tag('TAG3', pattern_tags))^0
+local h4 = lex:tag(lexer.PREPROCESSOR, pattern_h4) * (agenda_tags * lpeg.space)^0 * (lex:tag('PRIORITY4', pattern_priority) * lpeg.space)^0 * part_h4 * (lex:tag('TAG4', pattern_tags))^0
+local h5 = lex:tag(lexer.CONSTANT, pattern_h5) *     (agenda_tags * lpeg.space)^0 * (lex:tag('PRIORITY5', pattern_priority) * lpeg.space)^0 * part_h5 * (lex:tag('TAG5', pattern_tags))^0
```
orbitalquark: None of these headings appear to be used. Can they be deleted?

lexers/org.lua
```diff
@@ -0,0 +1,137 @@
...
+
+-- Links.
+local orgmode_link = '[[' * (lexer.nonnewline - ' ' - ']')^1 * ']' * ('[' * (lexer.nonnewline - ']')^1 * ']')^0 * ']'
+lex:add_rule('link', lex:tag(lexer.FUNCTION, orgmode_link))
```
orbitalquark: Please use the `lexer.LINK` tag.

lexers/org.lua
```diff
@@ -0,0 +1,137 @@
...
+
+-- Strings.
+-- local sq_str = P('L')^-1 * lexer.range("'")
+local dq_str = P('L')^-1 * lexer.range('"')
+-- lex:add_rule('string', lex:tag(lexer.STRING, sq_str + dq_str))
+lex:add_rule('string', lex:tag(lexer.STRING, dq_str))
```
orbitalquark: I think for simplicity you can use use

```
lex:add_rule('string', lex:tag(lexer.STRING, P('L')^-1 * lexer.range('"')))
```

lexers/org.lua
```diff
@@ -0,0 +1,137 @@
...
+
+-- Comments.
+lex:add_rule('comment', lex:tag(lexer.COMMENT,
+	lexer.starts_line(lexer.to_eol('# ')) +
+	lexer.range(lexer.starts_line('#+BEGIN_COMMENT'),
+		lexer.starts_line('#+END_COMMENT'))))
```
orbitalquark: Your block comments that start on a line will be highlighted as line comments. You want to match block comments first. I'd recommend using two variables: `line_comment` and `block_comment`, then have the rule match `block_comment + line_comment`.

orbitalquark (changes requested):
Thanks for taking the time to submit this! I've left some comments for work left to do.

lexers/org.lua
mcepl: Done, the same for `underline`.

lexers/org.lua
mcepl: Done

lexers/org.lua
mcepl: Absolutely, that’s from the original file, which however I have rewritten so much, I feel free to relicense to `Released under the conditions described in the LICENSE file.`

lexers/org.lua
mcepl: Done

lexers/org.lua
mcepl: done

lexers/org.lua
mcepl: Well, with my first name being “Matěj” I am interested in that `locale` part as well. For example, `alpha` should be slightly more than `R('AZ', 'az')`, shouldn’t it?

lexers/org.lua
mcepl: Thank you, good point!

lexers/org.lua
mcepl: DOne

lexers/org.lua
mcepl: Removed for now, the whole heading thing must be rewritten later.

lexers/org.lua
mcepl: Whole headings stuff copied from Markdown lexer.

lexers/org.lua
mcepl: Nonsense, I am using just `space`, there is no meaningful localization there.

fischerling:
> > `vis:win:draw()` does redraw the current window and thus relexes the viewport. This is what `vis-spellcheck` does when invoking one of its commands.
> 
> Does it? Any idea what’s wrong here then?

Pretty sure it does.

> ```lua
> vis.events.subscribe(vis.events.WIN_OPEN, function(win)
> 	-- Your per window configuration options e.g.
> 	if win.syntax == 'org' then
> 		vis:info("Modifying org lexer")
> 		local lex = vis.lexers.load(win.syntax)
> 		local nonnewline = vis.lexers.nonnewline
> 		local suse_abbrev = lpeg.B(lpeg.space) * ( lpeg.lower^-4 * lpeg.S("#-") )^1 * lpeg.digit^1 * #lpeg.space
> 		local orgmode_link = lpeg.P('[[') * (nonnewline - ' ' - ']')^1 * ']' * ('[' * (nonnewline - ']')^1 * ']')^0 * ']'
> 		lex:modify_rule('link', lex:tag(vis.lexers.FUNCTION, orgmode_link + suse_abbrev))
> 		win:draw()
> 	end
> end)
> ```

The problem with your code is, that you do not request the cached version of the lexer, and thus modify a copy that is not used by vis.

See: https://github.com/martanne/vis/blob/4970df09cea1e427b36469cc66340906cb5a0cb1/lua/vis.lua#L132

mcepl:
> The problem with your code is, that you do not request the cached version of the lexer, and thus modify a copy that is not used by vis.

Thank you! I needed to change this one line to make it work:

```lua
		local lex = vis.lexers.load(win.syntax, nil, true)
```

mcepl:
Fixed squashing, now it should be correct.

lexers/org.lua
```diff
@@ -0,0 +1,124 @@
...
+-- Font formats.
+lex:add_rule('bold', lex:tag('BOLD',      '*' * lexer.word * '*'))
+lex:add_rule('italic', lex:tag('ITALIC',    '/' * lexer.word * '/'))
+lex:add_rule('underline', lex:tag('UNDERLINE', '_' * lexer.alnum * '_'))
+
+-- ToDos.
+lex:add_rule('todo', lex:tag('TODO', lex:word_match('TODO')))
+lex:add_rule('done', lex:tag('DONE', lex:word_match('DONE')))
+lex:add_rule('wontfix', lex:tag('WONTFIX', lex:word_match('WONTFIX')))
+local agenda_tags = P('TODO') + 'DONE' + 'WONTFIX'
```
orbitalquark: This is unused now and can be removed.

lexers/org.lua
```diff
@@ -0,0 +1,124 @@
...
+
+-- DateTime.
+local DD =   lexer.digit * lexer.digit
+local date = DD * DD * '-' * DD * '-' * DD
```
orbitalquark: This is also unused, and can be removed.

lexers/org.lua
```diff
@@ -0,0 +1,124 @@
...
+-- * -- Headings.
+-- * local h1 = lex:tag(lexer.CLASS, pattern_h1) *        (agenda_tags * lpeg.space)^0 * (lex:tag('PRIORITY1', pattern_priority) * lpeg.space)^0 * part_h1 * (lex:tag('TAG1', pattern_tags))^0
+-- * local h2 = lex:tag(lexer.STRING, pattern_h2) *       (agenda_tags * lpeg.space)^0 * (lex:tag('PRIORITY2', pattern_priority) * lpeg.space)^0 * part_h2 * (lex:tag('TAG2', pattern_tags))^0
+-- * local h3 = lex:tag(lexer.FUNCTION, pattern_h3) *     (agenda_tags * lpeg.space)^0 * (lex:tag('PRIORITY3', pattern_priority) * lpeg.space)^0 * part_h3 * (lex:tag('TAG3', pattern_tags))^0
+-- * local h4 = lex:tag(lexer.PREPROCESSOR, pattern_h4) * (agenda_tags * lpeg.space)^0 * (lex:tag('PRIORITY4', pattern_priority) * lpeg.space)^0 * part_h4 * (lex:tag('TAG4', pattern_tags))^0
+-- * local h5 = lex:tag(lexer.CONSTANT, pattern_h5) *     (agenda_tags * lpeg.space)^0 * (lex:tag('PRIORITY5', pattern_priority) * lpeg.space)^0 * part_h5 * (lex:tag('TAG5', pattern_tags))^0
+
+-- Links.
+local orgmode_link = '[[' * (lexer.nonnewline - ' ' - ']')^1 * ']' * ('[' * (lexer.nonnewline - ']')^1 * ']')^0 * ']'
```
orbitalquark, line 62: I think `('[' * (lexer.nonnewline - ']')^1 * ']')` can be simplified to `lexer.range('[', ']', true)`

lexers/org.lua
```diff
@@ -0,0 +1,124 @@
...
+lex:add_rule('link', lex:tag(lexer.LINK, orgmode_link))
+
+-- Strings.
+lex:add_rule('string', lex:tag(lexer.STRING, P('L')^-1 * lexer.range('"')))
+
+-- Comments.
+local line_comment = lexer.starts_line(lexer.to_eol('# '))
```
orbitalquark, line 69: Is the space after '#' required? If not, then it can be removed.

orbitalquark (changes requested):
This looks nearly complete. Just some small things to double-check and one potential simplification.

lexers/org.lua
mcepl: It is slightly more complicated, [the link](https://orgmode.org/guide/Hyperlinks.html) looks like this:

```org
[[https://orgmode.org/][Org-mode (visible text)]]
```

lexers/org.lua
mcepl: [Yes](https://orgmode.org/guide/Comment-Lines.html):

> Lines starting with zero or more whitespace characters followed by one ‘#’ and a whitespace are treated as comments

orbitalquark:
Thanks for all of your work on this! Committed in https://github.com/orbitalquark/scintillua/commit/5e1cec89e2e0491f61d915e8ddb49a6c3b8ebbad.

PR #138, opened 2025-01-28
standardize definition of lexer and lex, remove whitespace definition

mcepl:
This should be mostly mechanical conversion, should be non-controversial.


mcepl:
OK, closing this one, and I will send patches per each lexer. Eventually.

PR #140, opened 2025-01-29
fix: modernize rest lexer

mcepl:

### DIFF
--- a/lexers/rest.lua
+++ b/lexers/rest.lua
@@ -1,11 +1,11 @@
 -- Copyright 2006-2025 Mitchell. See LICENSE.
 -- reStructuredText LPeg lexer.
 
-local lexer = require('lexer')
-local token, word_match, starts_line = lexer.token, lexer.word_match, lexer.starts_line
+local lexer = lexer
+local token, starts_line = lexer.token, lexer.starts_line
 local P, S = lpeg.P, lpeg.S
 
-local lex = lexer.new('rest')
+local lex = lexer.new(...)
 
 -- Literal block.
 local block = '::' * (lexer.newline + -1) * function(input, index)
@@ -19,8 +19,7 @@ local block = '::' * (lexer.newline + -1) * function(input, index)
 	end
 	return #input + 1
 end
-lex:add_rule('literal_block', token('literal_block', block))
-lex:add_style('literal_block', lexer.styles.embedded .. {eolfilled = true})
+lex:add_rule('literal_block', lex:tag('literal_block', block))
 
 -- Lists.
 local option_word = lexer.alnum * (lexer.alnum + '-')^0
@@ -31,7 +30,7 @@ local bullet_list = S('*+-') -- TODO: '•‣⁃', as lpeg does not support UTF-
 local enum_list = P('(')^-1 * (lexer.digit^1 + S('ivxlcmIVXLCM')^1 + lexer.alnum + '#') * S('.)')
 local field_list = ':' * (lexer.any - ':')^1 * P(':')^-1
 lex:add_rule('list', #(lexer.space^0 * (S('*+-:/') + enum_list)) *
-	starts_line(token(lexer.LIST,
+	starts_line(lex:tag(lexer.LIST,
 		lexer.space^0 * (option_list + bullet_list + enum_list + field_list) * lexer.space)))
 
 local any_indent = S(' \t')^0
@@ -40,15 +39,12 @@ local prefix = any_indent * '.. '
 
 -- Explicit markup blocks.
 local footnote_label = '[' * (lexer.digit^1 + '#' * word^-1 + '*') * ']'
-local footnote = token('footnote_block', prefix * footnote_label * lexer.space)
+local footnote = lex:tag('footnote_block', prefix * footnote_label * lexer.space)
 local citation_label = '[' * word * ']'
-local citation = token('citation_block', prefix * citation_label * lexer.space)
-local link = token('link_block', prefix * '_' *
+local citation = lex:tag('citation_block', prefix * citation_label * lexer.space)
+local link = lex:tag('link_block', prefix * '_' *
 	(lexer.range('`') + (P('\\') * 1 + lexer.nonnewline - ':')^1) * ':' * lexer.space)
 lex:add_rule('markup_block', #prefix * starts_line(footnote + citation + link))
-lex:add_style('footnote_block', lexer.styles.label)
-lex:add_style('citation_block', lexer.styles.label)
-lex:add_style('link_block', lexer.styles.label)
 
 -- Sphinx code block.
 local indented_block = function(input, index)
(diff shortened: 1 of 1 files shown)


lexers/rest.lua
```diff
@@ -19,8 +19,7 @@ local block = '::' * (lexer.newline + -1) * function(input, index)
 	end
 	return #input + 1
 end
-lex:add_rule('literal_block', token('literal_block', block))
-lex:add_style('literal_block', lexer.styles.embedded .. {eolfilled = true})
+lex:add_rule('literal_block', lex:tag('literal_block', block))
```
orbitalquark: We either need to use a known tag name or define the style for this tag in themes. Otherwise, editors will not know about it and cannot style it.

lexers/rest.lua
```diff
@@ -40,15 +39,12 @@ local prefix = any_indent * '.. '
 
 -- Explicit markup blocks.
 local footnote_label = '[' * (lexer.digit^1 + '#' * word^-1 + '*') * ']'
-local footnote = token('footnote_block', prefix * footnote_label * lexer.space)
+local footnote = lex:tag('footnote_block', prefix * footnote_label * lexer.space)
 local citation_label = '[' * word * ']'
-local citation = token('citation_block', prefix * citation_label * lexer.space)
-local link = token('link_block', prefix * '_' *
+local citation = lex:tag('citation_block', prefix * citation_label * lexer.space)
+local link = lex:tag('link_block', prefix * '_' *
```
orbitalquark: The styles for these tags needs to be defined in themes. Perhaps something like `lexer.REFERENCE .. '.footnote'`, `lexer.REFERENCE .. '.citation'`, etc.

lexers/rest.lua
```diff
@@ -61,60 +57,24 @@ local indented_block = function(input, index)
 end
 local code_block =
 	prefix * 'code-block::' * S(' \t')^1 * lexer.nonnewline^0 * (lexer.newline + -1) * indented_block
-lex:add_rule('code_block', #prefix * token('code_block', starts_line(code_block)))
-lex:add_style('code_block', lexer.styles.embedded .. {eolfilled = true})
+lex:add_rule('code_block', #prefix * lex:tag('code_block', starts_line(code_block)))
```
orbitalquark: I think we can use `lexer.CODE` for the tag.

lexers/rest.lua
```diff
@@ -61,60 +57,24 @@ local indented_block = function(input, index)
...
-	'centered', 'hlist', 'glossary', 'productionlist',
-	-- Showing code examples.
-	'highlight', 'literalinclude',
-	-- Miscellaneous
-	'sectionauthor', 'index', 'only', 'tabularcolumns'
-} * '::' * lexer.space)
-local unknown_directive = token('unknown_directive', prefix * word * '::' * lexer.space)
+local known_directive = lex:tag('directive', prefix * lex:word_match('directive') * '::' * lexer.space)
+local sphinx_directive = lex:tag('sphinx_directive', prefix * lex:word_match('sphinx_directive') * '::' * lexer.space)
+local unknown_directive = lex:tag('unknown_directive', prefix * word * '::' * lexer.space)
```
orbitalquark: Again, we need known tags or define styles in themes. The same idea applies to the rest of the custom tag names in this lexer.

lexers/rest.lua
```diff
@@ -135,38 +95,32 @@ local underline = lpeg.Cmt(starts_line(adornment), function(_, index, adm, c)
...
 end)
 -- Token needs to be a predefined one in order for folder to work.
-lex:add_rule('title', token(lexer.HEADING, overline + underline))
+lex:add_rule('title', lex:tag(lexer.HEADING, overline + underline))
 
 -- Line block.
-lex:add_rule('line_block_char', token(lexer.OPERATOR, starts_line(any_indent * '|')))
-
--- Whitespace.
-lex:add_rule('whitespace', token(lexer.WHITESPACE, S(' \t')^1 + lexer.newline^1))
```
orbitalquark: The migration guide allows for lexers to define their own whitespace rules if whitespace is significant. That is the case here. However, reST is a special case in that its whitespace rule is below other rules (normally it's at the top, like the Markdown lexer). I honestly don't know how this is going to work in a migrated lexer.

Regardless, removing this rule will break the lexer. We need it, I'm just not sure how.

orbitalquark (changes requested):
Honestly, this is probably my most dreaded lexer on the migration list. There are a lot of things that need to be considered. I've left some comments illustrating some of them. If you want to abandon the effort on this one, I do not blame you. A lot of testing is going to be involved.

lexers/rest.lua
mcepl: Readded

lexers/rest.lua
mcepl: I would say that this is more `lexer.LABEL` for the body of a note, whereas `lexer.REFERENCE` is for the link to it.

lexers/rest.lua
mcepl: Done.

mcepl:
I agree, this is not an easy lexer to deal with. On the other hand, I use reStructuredText every day (being a Python developer, but I really prefer it to Markdown, whenever I can; https://matej.ceplovi.cz/blog/ and everything on that site has been written in rST), so I have a strong incentive to get it right.

I fully expect that we will be on this PR for some time before we get it better. Contributing now, the first round of my revisions. What do you think?

lexers/rest.lua
```diff
@@ -19,7 +19,7 @@ local block = '::' * (lexer.newline + -1) * function(input, index)
 	end
 	return #input + 1
 end
-lex:add_rule('literal_block', lex:tag('literal_block', block))
+lex:add_rule(lexer.LABEL .. 'literal', lex:tag(lexer.LABEL .. 'literal', block))
```
orbitalquark: I suggest `lex:add_rule('literal_block', lex:tag(lexer.LABEL .. '.literal', block))`

Note that rule names can be simple strings, but tag names can be "subclassed" with dots. I added a dot. Please add dots in each of your subsequent tags.

orbitalquark (changes requested):
Thanks for letting me know that you are actively using this lexer and have an interest in getting it right. That makes me more comfortable with any upcoming changes. I've commented that your tags need a '.' Other than that, I am okay with this first round of revisions.

lexers/rest.lua
mcepl: Is it that? (and yes, this needs to be mergre-squashed)

lexers/rest.lua
```diff
@@ -135,38 +95,36 @@ local underline = lpeg.Cmt(starts_line(adornment), function(_, index, adm, c)
...
 	(strong + em + inline_literal + link_ref + interpreted + footnote_ref + citation_ref +
 		substitution_ref + link) * -lexer.alnum)
-lex:add_style('inline_literal', lexer.styles.embedded)
-lex:add_style('role', lexer.styles.class)
-lex:add_style('interpreted', lexer.styles.string)
 
 -- Other.
-lex:add_rule('non_space', token(lexer.DEFAULT, lexer.alnum * (lexer.any - lexer.space)^0))
-lex:add_rule('escape', token(lexer.DEFAULT, '\\' * lexer.any))
+-- TODO Not sure what it is good for
```
orbitalquark: This prevents matching any of the above rules in part of a word. For example, `a::block` would not tag `::block` as a block. The entire entity would be tagged `lexer.DEFAULT`.

mcepl: Yes, I tried to comment these out and the result was a disaster, thank you.

orbitalquark:
Thanks for your work on this. I've committed it in https://github.com/orbitalquark/scintillua/commit/9674904c379bf37f1014defc3f8577d2edb39cc6 with the one change of tagging directives as keywords instead of operators. Scintillua's themes do not style operators as anything special, so directives would not be visually distinctive. I hope this is okay with you.

I also needed to update Scintillua's themes to reflect the tags you've chosen.

mcepl:
On Sat Apr 19, 2025 at 5:56 PM CEST, orbitalquark wrote:
> Thanks for your work on this. I've committed it in https://github.com/orbitalquark/scintillua/commit/9674904c379bf37f1014defc3f8577d2edb39cc6 with the one change of tagging directives as keywords instead of operators. Scintillua's themes do not style operators as anything special, so directives would not be visually distinctive. I hope this is okay with you.

No problem with me, I will have to just adjust our themes accordingly.

PR #144, opened 2025-02-08
Proposed SIL TeX-like lexer (SILE)

Omikhleia:
Greetings,

This PR proposes a Scintillua lexer for the "SIL in TeX-like flavor" syntax used by the [SILE typesetting system](https://sile-typesetter.org/).

It's my first attempt a (theoretically simple) lexer, I hope I did it correctly ;)

Regards.

### DIFF
--- a/lexers/sil.lua
+++ b/lexers/sil.lua
@@ -1,0 +1,180 @@
+-- Copyright 2025, Didier Willis
+-- (Donated to Scintillua, under the same license.)
+-- See LICENSE.
+--
+-- Lexer for the SIL "TeX-like" syntax used by the SILE typesetting system.
+--
+-- The SILE typesetting system (https://sile-typesetter.org/) can support
+-- different input formats, via dedicated "inputters".
+-- It is not tied to a given input syntax.
+-- Yet, the "SIL in TeX-like flavor" format is a common one, included in the
+-- core distribution and used in most of the examples and documentation.
+--
+-- SIL TeX-like is similar to LaTeX, but with a parity with SIL XML.
+-- I.e. \foo[attr=val]{...} is equivalent to <foo attr="val">...</foo>.
+-- Environments are syntactic sugar for commands.
+-- I.e. \begin[attr=val]{foo}...\end{foo} is the same as \foo[attr=val]{...}.
+-- Rules below try to respect this parity.
+
+local lexer = lexer
+local P, S, R, Cb, Cg, Ct, Cmt = lpeg.P, lpeg.S, lpeg.R, lpeg.Cb, lpeg.Cg, lpeg.Ct, lpeg.Cmt
+
+local lex = lexer.new(...)
+local ws = lex:get_rule('whitespace')
+
+-- 1. Syntax bits.
+
+-- SIL identifiers can contain letters, numbers, and the characters :-.
+-- There are additional rules (e.g. no leading digits), but we'll keep it simple here.
+local identifier = lexer.alnum^1 * (S(':-') * lexer.alnum^1)^0
+
+-- Reserved hard-coded "pass-through" commands/environments.
+local reserved_specials = {
+	ftl = 'text', -- Well Fluent has a syntax, but let's not care here.
+	lua = 'lua',
+	math = 'tex', -- We'd need a (La)TeX math-only lexer to handle this properly.
+	raw = 'text',
+	script = 'lua',
+	-- sil = ... -- It's the default here, so no need to add a rule for it.
+	xml = 'xml',
+	use = 'lua',
+}
+-- Other reserved keywords are "comment" and "begin"/"end",
+-- but we'll handle them in the rules below.
+
+-- Parameters (key-value pairs).
+local eq = lex:tag(lexer.OPERATOR, '=')
+local simple_value = (P(1) - S(',;]'))^1
+local quoted_value = lexer.range('"', false, false)
+local param = lex:tag(lexer.ATTRIBUTE, identifier) * eq * lex:tag(lexer.STRING, quoted_value + simple_value)
+local param_list = param * (ws^0 * lex:tag(lexer.OPERATOR, ',') * ws^0 * param)^0
+local optparams = (lex:tag(lexer.OPERATOR, '[') * ws^0 * param_list^0 * ws^0 * lex:tag(lexer.OPERATOR, ']'))^0
+
+-- 2. Comments.
+local line_comment = lexer.to_eol('%')
+local env_comment = lexer.range(P('\\begin') * optparams * P('{comment}'), P('\\end{comment}'))
+local cmd_comment = P('\\comment') * optparams * lexer.range('{', '}', false, false, true)
+lex:add_rule('comment', lex:tag(lexer.COMMENT, line_comment + env_comment + cmd_comment))
+
+-- 3. Special reserved pass-through commands/environments.
+
+local function check_exit_brace_level(_, _, current_level)
+	current_level = tonumber(current_level)
+	return current_level == 0
+end
+
+local function increment_brace_level(increment)
+	local function update_brace_level(_, _, current_level)
+		current_level = tonumber(current_level)
+		local next_level = tostring(current_level + increment)
+		return true, next_level
+	end
+	return Cg(Cmt(Cb('brace_level'), update_brace_level), 'brace_level')
+end
+
+local is_exit_brace = Cmt(Cb('brace_level'), check_exit_brace_level)
+local init_brace_level = Cg(Ct('') / '0', 'brace_level')
+
+for name, lang in pairs(reserved_specials) do
+	-- Order matters: environments, commands with arguments, commands without arguments.
+	-- We need alt names for multiple embeddings and rules.
+	local base_rule_id = name .. '_' .. lang
+
+	-- 3.1. Reserved environments.
+	-- Ex. \begin{lua} ... Lua code ... \end{lua}
+	local env_embedder = lexer.load(lang, base_rule_id .. '_env')
+	lex:embed(
+		env_embedder,
+		lex:tag(lexer.FUNCTION_BUILTIN, '\\begin') * optparams
+			* lex:tag(lexer.OPERATOR, '{') * lex:tag(lexer.FUNCTION_BUILTIN, name) * lex:tag(lexer.OPERATOR, '}'),
+		lex:tag(lexer.FUNCTION_BUILTIN, '\\end')
+			* lex:tag(lexer.OPERATOR, '{') * lex:tag(lexer.FUNCTION_BUILTIN, name) * lex:tag(lexer.OPERATOR, '}'))
+
+	-- 3.2. Reserved commands.
+	-- Ex. \lua{... Lua code ...}
+	-- The hard trick here is that we want to want to keep track of the paired braces,
+	-- in order to exit the embedding on the right closing brace.
+	local cmd_embedder = lang == 'text'
+		and lexer.new(base_rule_id .. '_cmd') -- pseudo-lexer for text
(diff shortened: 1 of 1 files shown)


Omikhleia:
N.B. An illustration of what it does: https://github.com/sile-typesetter/sile/discussions/2220#discussioncomment-12105274

lexers/sil.lua
```diff
@@ -0,0 +1,106 @@
...
+}
+-- Other reserved keywords are "comment" and "begin"/"end"
+-- But we'll handle them in the rules below.
+
+-- Parameters (key-value pairs).
+local eq = lex:tag(lexer.OPERATOR, '=')
+local simple_value = (P(1) - S(',;]'))^1
+local quoted_value = lexer.range('"', false, false)
+local param = lex:tag(lexer.ATTRIBUTE, identifier) * eq * lex:tag(lexer.STRING, quoted_value + simple_value)
+local param_list = param * (ws^0 * lex:tag(lexer.OPERATOR, ',') * ws^0 * param)^0
```
orbitalquark, line 50: Using `ws` will include newlines. I don't know if that's acceptable syntax, but some editors may take issue when starting a lex at the beginning of a line that is inside a `param_list`. Scintillua cannot backtrack far enough to the start of the list. I'd recommend declaring `local ws = lex:tag(lexer.WHITESPACE, S(' \t'))` so that only horizontal space matches.

lexers/sil.lua
```diff
@@ -0,0 +1,106 @@
...
+local optparams = (lex:tag(lexer.OPERATOR, '[') * ws^0 * param_list^0 * ws^0 * lex:tag(lexer.OPERATOR, ']'))^0
+
+-- 2. Comments.
+local line_comment = lexer.to_eol('%')
+local env_comment = lexer.range(P('\\begin') * optparams * P('{comment}'), P('\\end{comment}'))
```
orbitalquark, line 55: Typically Scintillua lexers will allow the trailing delimiter to be optional (`P('\end{comment}')^-1`), so that the comment is highlighted correctly as you type. I'd recommend doing that here.

lexers/sil.lua
```diff
@@ -0,0 +1,106 @@
...
+local cmd_comment = lexer.range('\\comment{', '}')
```
orbitalquark: Do you mind changing this to `'\\comment' * lexer.range('{', '}')`?

lexers/sil.lua
```diff
@@ -0,0 +1,106 @@
...
+lex:add_rule('comment', lex:tag(lexer.COMMENT, line_comment + env_comment + cmd_comment))
+
+-- 3. Special reserved pass-through commands/environments.
+for name, lang in pairs(reserved_specials) do
+	-- Order matters: environments, commands with arguments, commands without arguments
+	-- We need alt names for multiple embeddings and rules
+	local base_rule_id = name .. lang
+	local embedder = lexer.load(lang, base_rule_id .. '_cmd')
+	lex:embed(
```
orbitalquark: This is clever!

lexers/sil.lua
```diff
@@ -0,0 +1,106 @@
...
+		lex:tag(lexer.FUNCTION_BUILTIN, '\\end')
+			* lex:tag(lexer.OPERATOR, '{') * lex:tag(lexer.FUNCTION_BUILTIN, name) * lex:tag(lexer.OPERATOR, '}'))
+	lex:add_rule(base_rule_id .. '_cmd_no_arg', lex:tag(lexer.FUNCTION_BUILTIN, P('\\' .. name)) * optparams)
+end
+
+-- 4. Sections (for mere convenience / visibility).
+-- As of 0.15.9 SILE's default book class has chapter, section, subsection.
+-- 3rd-party module resilient.sile adds part, appendix, subsubsection, and frontmatter, mainmatter, backmatter.
+lex:add_rule('section', lex:tag('command.section', '\\' *
+	word_match('frontmatter mainmatter backmatter part chapter appendix section subsection subsubsection'))
```
orbitalquark: Please use a `lex:word_match()` and `lex:set_word_list()` pair, like with other lexers.

orbitalquark (changes requested):
Thanks for your contribution! This looks pretty good. I've left some comments for some things to address.

lexers/sil.lua
Omikhleia: :blush:

lexers/sil.lua
Omikhleia: Done in fixup (I'll rebase --autosquash when all is ok)

lexers/sil.lua
Omikhleia: Done in fixup (I'll rebase --autosquash when all is ok)
This however wasn't robust, so I went a bit further in the rewrite:
 - I forgot to take into account the optional params (seldom used in such a construct, but accepted)
 - The { } needs to be balanced: `\comment{Some \em{stuff} commented}` is valid...

lexers/sil.lua
Omikhleia: Not changed: I tried, but with the `^-1` the end was never satisfied (when present) ??

lexers/sil.lua
Omikhleia: Newlines are accepted there, and are not that uncommon. E.g.
```
\framebox[
          bordercolor=red, fillcollor=#f0f0f0,
          shadow=true, shadowcolor=black
]{Some box}
```
(I tend to do this myself, for separating parameters of different nature)

lexers/sil.lua
Omikhleia: Not that clever actually, and there's an edge case for commands that I'have not idea how to address.
This would be valid:
```
\lua{SILE.process({ "text" })}
```
So the "end rule" for embedding cannot just be a }, it has to be balanced... I haven't found such a case in the sources (most embedded lexers have clear start/end rules not expected in child lexer) ?

orbitalquark: Sorry for the delayed response.

I don't think there is an easy way to address this. The only thing I can think of is when encountering a '{' or '}', increase or decrease a count variable, and store that in a persistent lexer line state. However, I'm not sure that would work for the one-line example you gave.

I think you'll have to accept the limitation, or not perform this kind of highlighting.

Omikhleia: > Sorry for the delayed response. The only thing I can think of is when encountering a '{' or '}', increase or decrease a count variable (...)

Thanks for you feedback.
I've added in a fixup some lpeg-heavy machinery to keep track of the brace pairing level and exit the child lexer upon the first unpaired closing brace. The "state" is kept in the parsing patterns via a table.[^1]

It does work well, as far as I can tell, for syntax highlighting a whole file:
image

(with ugly colors here to better distinguish the items :smile_cat: )

But I don't know how an _editor_ (SciTE, I suppose) behaves in those case when characters are entered by the user as he is typing).

[^1]: Logic somewhat inspired / derived from the way the lunamark library keeps track of nested divs in Markdown, with a similar technique.

Omikhleia: N.B. Of course, the pairing and child lexer hacking is fairly naive: there would still probably be an issue with unpaired braces part of a comment, escape sequence or other acceptable syntax (e.g. a string, etc.) in the embedded child language. 
But I think this is such an edge case that it's probably acceptable (and several other editor solutions do not handle these cases very gracefully either, anyway...)

orbitalquark (changes requested):
I've played around with your lexer, and I know you've put a lot of work into this, but it's too brittle. It'll get the parse right when opening a file, but once you start typing into it, it all falls apart.

For example, I've opened this file:

```
\begin[papersize=a5]{document} 
\nofolios 
\neverindent 
\use[module=packages.math] 

\chapter{Some SIL code} 

\section[numbered=false]{Some SIL code} 

\begin[rule="0.4pt"]{epigraph} 
Lorem ipsum dolor sit amet, consectetur adipiscing elit.

\comment{A comment} 

\ftl{Some text} 

\xml{<tag>Some XML code</tag>} 

\em{Sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.} 
\caption{John \font[variant=+smcp]{Doe}.} 
\end{epigraph} 

\math[mode=diplay]{ 
	\int_{0)^{1} x^2 dx = \frac{1}{3} 
}

\lua{SILE.typesetter:typeset("Hello, world!")}

\font[family=DejaVu Serif, size=12pt]{Hello, world!} 

\font[family=DejaVu Serif, size=12pt]

\begin{lua} 
SILE.typesetter:typeset("Hello, world!") 
\end{lua}

\end{document}
```

I put my caret on the `\xml{<tag>Some XML code</tag>}` line after the first `{` and hit enter. My editor will start lexing the line starting with `<tag>` (the previous line is `\xml{`). I get an error:

```
/path/to/lexers/lexer.lua:1255: back reference 'brace_level' not found
```

If I undo the change, now pressing enter elsewhere in the file gives me the same error; editing becomes impossible. In fact, pressing enter anywhere in the file after immediately opening it gives me the error.

---

I also tried a simpler case:

```
\framebox[bordercolor=red, fillcollor=#f0f0f0, shadow=true, shadowcolor=black]{Some box}
```

This highlights great, but now I want to split it into multiple lines between the `[ ]`. After I press enter after the first `[` followed by Tab (for indentation), highlighting immediately breaks, because lexing starts at `bordercolor`, which loses the context from what is now the previous line:

image

---

In my experience, lexers need to be written as context-free as possible in order to highlight reliably when editing. For example, the HTML lexer is aware that tag attributes may span multiple lines, so it only highlights an attribute if an `=` immediately follows it: https://github.com/orbitalquark/scintillua/blob/6665a6adddeaf0c90496929d916ecdf59f131b19/lexers/html.lua#L47-L48 Similarly it only highlights numbers if they trail a `=`, indicating an attribute value. Yes, it's not perfect, but it gets it right a vast majority of the time and it's robust for editing.

---

What you have here is an impressive one-time lexer. You've done some clever things I didn't realize could be viable. However, Scintillua's lexers are designed to lex files being edited, so we need to keep that in mind.

I'm not sure what to suggest moving forward. Maybe just take a step back and see what you can do to make it a bit more context-free. I would also encourage you to try editing code with your lexer active. You said you didn't know how an editor like SciTE behaves as the user types. Experiment and see :) You can iron out bugs and issues as they arise.

Omikhleia:
> I've played around with your lexer, and I know you've put a lot of work into this, but it's too brittle. 
> (...)
> Experiment and see :)

Thanks a lot for all the feedbacks and hints, @orbitalquark 
I Just installed Textadept, and indeed reproduced the same problems... Ouch indeed.

> What you have here is an impressive one-time lexer. You've done some clever things I didn't realize could be viable. However, Scintillua's lexers are designed to lex files being edited, so we need to keep that in mind.
> I'm not sure what to suggest moving forward.

I don't know either (I experimented some other ways towards a more context-free approach tonight, but without success so far, and my free time is running short, riding two horses with one behind...)

Well... it somewhat works for highlighting a "complete file", so for now:
 - I'll use that code as an "addon" in my own use case (the [highlighter.sile](https://github.com/Omikhleia/highlighter.sile) module for SILE, which does take a complete input so doesn't have to handle the issues of a typist editing a file on the fly, a "one-time" lexer is sufficient there...).
 - I've just switched this PR to Draft status. It's kind of a step back which I didn't foresee, but thanks again for the encouragements and ideas. I'm glad I discovered this repository, and now Textadept -- The effort was certainly not in vain, I've learnt a few things in that adventure, and I hope to return to it at some later point. (I'm also considering another lexer, which I hope is _even_ simpler... though I discovered here that the devil is sometimes in the details, but _heh,_ at least, that one would have no need for embedded child languages and might have a context-free parsing... We'll see... _Once bitten, twice shy_ :cat: ).

orbitalquark:
I think you have a great attitude that will carry you to success. There's no hurry, so take your time :) Cheers.

PR #145, opened 2025-02-18
Migrate OCaml Lexer

samuelmarquis:
This one was pretty weird.

My specific qualm and the reason for doing this is that being able to accurately lex single-quoted strings in OCaml would be kind of a nightmare. You very frequently have type annotations like `'a` and match on things called `a'`. As far as I can tell, there's not a fantastic way to handle both of those things and still come out with something that isn't nightmarish in its complexity--so IMO, the best way to do it is to just completely nuke that.

I also added `list` as a type. There's definitely a lot more types that could be added, but I think `list` is probably pervasive enough to be the absolute bare minimum.

It's been a while since I've written one of these; let me know if I made any silly mistakes.

### DIFF
--- a/lexers/caml.lua
+++ b/lexers/caml.lua
@@ -1,30 +1,55 @@
 -- Copyright 2006-2025 Mitchell. See LICENSE.
 -- OCaml LPeg lexer.
+-- Migrated by Samuel Marquis.
 
-local lexer = require('lexer')
-local token, word_match = lexer.token, lexer.word_match
+local lexer = lexer
 local P, S = lpeg.P, lpeg.S
 
-local lex = lexer.new('caml')
-
--- Whitespace.
-lex:add_rule('whitespace', token(lexer.WHITESPACE, lexer.space^1))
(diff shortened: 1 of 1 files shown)


samuelmarquis:
Would also love to see this one renamed from `caml` to `ocaml` if you're set on fixing the names of things a la `ansi_c`

orbitalquark:
Thanks for taking the time to start on this! Does OCaml have methods and constants? Now is the time to "upgrade" the lexer to recognize things it didn't before. See the other migrated lexers for examples.

samuelmarquis:
I'm working on a compiler project in OCaml to learn the language right now--I will almost certainly be returning to this lexer over the next couple months.

There's a LOT of stuff in the standard library--but IMO, including much of that would be dubious because it's in a C++ sort of place where there's several different standard library replacements that many people seem to prefer. The language with no imports is pretty bare, and I don't think there's a standard syntax for constants.

Methods... it's probably called Objective CAML for a reason. I'll take a look.

PR #147, opened 2025-02-19
Add some fixes to Dart Lexer.

M0JXD:
A small PR for the Dart lexer that:
- Adds Function highlighting
- Adds import, async and await to the keywords list

### DIFF
--- a/lexers/dart.lua
+++ b/lexers/dart.lua
@@ -1,55 +1,68 @@
 -- Copyright 2013-2025 Mitchell. See LICENSE.
 -- Dart LPeg lexer.
 -- Written by Brian Schott (@Hackerpilot on Github).
+-- Migrated by Jamie Drinkell
 
-local lexer = require('lexer')
-local token, word_match = lexer.token, lexer.word_match
+local lexer = lexer
 local P, S = lpeg.P, lpeg.S
 
-local lex = lexer.new('dart')
-
--- Whitespace.
-lex:add_rule('whitespace', token(lexer.WHITESPACE, lexer.space^1))
+local lex = lexer.new(...)
 
 -- Keywords.
-lex:add_rule('keyword', token(lexer.KEYWORD, word_match{
-	'assert', 'break', 'case', 'catch', 'class', 'const', 'continue', 'default', 'do', 'else', 'enum',
-	'extends', 'false', 'final', 'finally', 'for', 'if', 'in', 'is', 'new', 'null', 'rethrow',
-	'return', 'super', 'switch', 'this', 'throw', 'true', 'try', 'var', 'void', 'while', 'with'
-}))
-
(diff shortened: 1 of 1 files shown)


orbitalquark:
Is this ready for review? If it is, please request one. I tend to not track WIP PRs until the owner marks it as ready.

M0JXD:
Hi,

These are just some quick fixes so I could make use of it but I think the Dart lexer still needs [modernized](https://github.com/orbitalquark/scintillua/issues/76). I think this lexer is based on Dart 2, when they added null safety in version 3 it changed the language considerably.

I want to have a look at doing this properly for Scintillua 6 against the Dart v3 language spec, but I'm not sure when. If you would like you can merge these for now as it's mostly usable against v3, but if you want to hold out for the job to be done properly that's no problem either as I think I'm the only person using Textadept for Dart :) (btw thanks for the code actions stuff in the LSP module, it works great)

Thank you

M0JXD:
Hi Mitchell,

I've modernized as per the guide and added a couple more fixes for Dart 3. There's still some things that would be nice to add (mainly Dart's [String interpolation](https://dart.dev/language/built-in-types#strings)) but that's not necessary for general use.

Thanks

orbitalquark:
Thanks. I've committed this with a change to the function rule to not tag as a function any whitespace between the function name and '('. https://github.com/orbitalquark/scintillua/commit/31323ceb7b4cbcc17fe9fadf353e472b91f923b5.

Feel free to close this if you're satisfied, or you can leave it open for any string interpolation work you want to tackle later.

M0JXD:
Thanks for committing and the functions rule fix. I'd like to tackle interpolation but not urgently - even the code samples on the linked page are not highlighted, and in several other editors it's not too. I can open a new PR when I get round to it.

However I'm comparing the the highlighting behaviour of a few keywords to a couple of other editors. The ones tagged as `lexer.CONSTANT` are keywords that in some contexts can be used as names, e.g. `static`. However in my experience it's infrequent that they get used outside of their keyword use, and in other editors like Xed and Kate they get highlighted as normal keywords/types with disregard for such context. So I can add them to the other lists and remove the constants list entirely if you're in agreement with changing that?

orbitalquark:
> Thanks for committing and the functions rule fix. I'd like to tackle interpolation but not urgently - even the code samples on the linked page are not highlighted, and in several other editors it's not too. I can open a new PR when I get round to it.
> 
> However I'm comparing the the highlighting behaviour of a few keywords to a couple of other editors. The ones tagged as `lexer.CONSTANT` are keywords that in some contexts can be used as names, e.g. `static`. However in my experience it's infrequent that they get used outside of their keyword use, and in other editors like Xed and Kate they get highlighted as normal keywords/types with disregard for such context. So I can add them to the other lists and remove the constants list entirely if you're in agreement with changing that?

Yes, I think they should be considered keywords.

PR #149, opened 2025-03-27
Adds better support for Typst

ghost:
This PR adds basic support for Typst. It's pretty basic for now and still has many areas to improve, since I'm still discovering LPeg,  I'd love to hear any thoughts or suggestions you have!!!
[Typst Reference](https://typst.app/docs/reference/syntax/)

Preview from vis editor

image

### DIFF
--- a/lexers/typst.lua
+++ b/lexers/typst.lua
@@ -93,6 +93,7 @@ end
 
 -- Keywords, functions... don't need '#' when in code
 -- the character `#` is not valid in code
+-- TODO: only enable styling and text related rules when in []
 add_rules(emb_lex, '')
 
 lex:embed(emb_lex, embed_start, embed_end)
--- a/lexers/typst.lua
+++ b/lexers/typst.lua
@@ -39,12 +39,13 @@ lex:add_rule('comment', lex:tag(lexer.COMMENT,
 lex:add_rule('list', lex:tag(lexer.LIST,
   lexer.starts_line(lexer.digit^1 * '.' + S('+-'), true) * S(' \t')))
 
--- TODO: The keywords currently are matched ANYWHERE (even in plain text), fix this later
+-- TODO: The keywords are currently matched only when precedded by #
+-- we must also match when nested in #{...}
 local keyword_condition = B('#') * lex:word_match(lexer.KEYWORD)
-lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD)))
+lex:add_rule('keyword', lex:tag(lexer.KEYWORD, keyword_condition))
 
--- NOTE: same goes for the functions and fields
-local func = lex:tag(lexer.FUNCTION, (B('.') + B('#')) * lexer.word * P('()'))
+-- FIXME: This mustn't apply when in plain text
+local func = lex:tag(lexer.FUNCTION, (B('.') + B('#')) * lexer.word * P('('))
 lex:add_rule('function', func)
 
 local field = lex:tag('FIELD', B('.') * lexer.word * (lexer.any - P('(')))
--- a/lexers/typst.lua
+++ b/lexers/typst.lua
@@ -4,64 +4,60 @@ local P, S, B = lpeg.P, lpeg.S, lpeg.B
 
 local lex = lexer.new(...)
 
-local hspace = lexer.space - '\n'
-local blank_line = '\n' * hspace^0 * ('\n' + P(-1))
-local punct_space = lexer.punct + lexer.space
+local horizontal_space = lexer.space - '\n'
 
-local function flanked_range(s, not_inword)
-  local fl_char = lexer.any - s - lexer.space
-  local left_fl = B(punct_space - s) * s * #fl_char + s * #(fl_char - lexer.punct)
-  local right_fl = B(lexer.punct) * s * #(punct_space - s) + B(fl_char) * s
-  return left_fl * (lexer.any - blank_line - (not_inword and s * #punct_space or s))^0 * right_fl
+local keyword_match = lex:word_match(lexer.KEYWORD) * #' '
+
+local bold_text = -B('\\') * lex:tag(lexer.BOLD, lexer.range('*', '*'))
+lex:add_rule('bold', bold_text)
+local italic_text = -B('\\') * lex:tag(lexer.ITALIC, lexer.range('_', '_'))
+lex:add_rule('italic', italic_text)
+
+
+local function header(level)
+  local equals_signs = P('=')^level
+  local standalone_header = lexer.starts_line(horizontal_space^0 * equals_signs * horizontal_space^1) * (lexer.any - S('#@<'))^0
+  local bracketed_header = P('[') * (lexer.space + '\n')^0 * equals_signs * horizontal_space^0 * (lexer.any - P(']'))^0 * (lexer.space + '\n')^0 * P(']')
+  return lex:tag(string.format('%s.h%s', lexer.HEADING, level), standalone_header + bracketed_header)
 end
+lex:add_rule('header', header(6) + header(5) + header(4) + header(3) + header(2) + header(1))
 
-local asterisk_strong = flanked_range('*', true)
-lex:add_rule('strong', lex:tag(lexer.BOLD, asterisk_strong))
+local label_definition = -B('\\') * lex:tag('LABEL', P('<') * lexer.word * P('>'))
+local label_call = -B('\\') * lex:tag('LABEL', P('@') * lexer.word)
+lex:add_rule('label', label_definition + label_call)
 
-local underscore_em = flanked_range('_', true)
-lex:add_rule('em', lex:tag(lexer.ITALIC, underscore_em))
-
-local function h(n)
-  local equals = P('=')^n
-  local standalone = lexer.starts_line(hspace^0 * equals * hspace^1) * lexer.to_eol(lexer.nonnewline)
-  local bracketed = P('[') * (lexer.space+'\n')^0 * equals * hspace^0 * (lexer.any - P(']'))^0 * (lexer.space+'\n')^0 * P(']')
-  return lex:tag(string.format('%s.h%s', lexer.HEADING, n), standalone + bracketed)
-end
-lex:add_rule('header', h(6) + h(5) + h(4) + h(3) + h(2) + h(1))
-
-local raw_str = lexer.range('`', false, false)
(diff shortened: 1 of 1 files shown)


orbitalquark:
Thanks for starting on this! When you think it's ready, request a review and I'll have a look. Also, please give a link to Typst documentation.

ghost:
Thanks, will surely do
 
i know this is out of context but i have a question about a rule i have:
```lua
local method = lex:tag("UNKNOWN", '#' * lexer.word * '.') * (lex:tag(lexer.FUNCTION_METHOD, lexer.word) * P('('))
lex:add_rule('func_method', method)
```
this is supposed to match the following expression:
`#dict.len()`
and it does, but I don't want to tag the `#dict.` as anything for now (not even `UNKNOWN`), but only use it to match the whole expr and only tag `len`, similar to non-input-consuming operators in lpeg, which when i try to use them the expression isn't matched anymore

ghost:
This is what I achieved as of now

2025-03-28_17-27-55_focused_window

some things are still not working correctly as I don't know enough LPeg to get them to work (like embedded parsing: notice the `if` (on the right pane) in the `#for` block not being tagged as keyword)

[Typst's syntax crate (includes lexer, parser...)](https://github.com/typst/typst/tree/main/crates/typst-syntax)

orbitalquark:
> Thanks, will surely do
> 
> i know this is out of context but i have a question about a rule i have:
> 
> ```lua
> local method = lex:tag("UNKNOWN", '#' * lexer.word * '.') * (lex:tag(lexer.FUNCTION_METHOD, lexer.word) * P('('))
> lex:add_rule('func_method', method)
> ```
> 
> this is supposed to match the following expression: `#dict.len()` and it does, but I don't want to tag the `#dict.` as anything for now (not even `UNKNOWN`), but only use it to match the whole expr and only tag `len`, similar to non-input-consuming operators in lpeg, which when i try to use them the expression isn't matched anymore

You always have to tag matched text with something in order to move on and tag stuff that follows. You can use the `lexer.DEFAULT` tag for the stuff you don't care about.

ghost:
@orbitalquark, I think the work is ready for a review. I added comments in the code about things I wasn't sure how to do, so I’d love your feedback on those and any other suggestions you have

lexers/typst.lua
```diff
@@ -0,0 +1,125 @@
+local lexer = require('lexer')
```
orbitalquark: Modern lexers use `local lexer = lexer` now.

lexers/typst.lua
```diff
@@ -0,0 +1,125 @@
...
+local token = lexer.token
```
orbitalquark: Modern lexers don't need this anymore.

lexers/typst.lua
```diff
@@ -0,0 +1,125 @@
...
+ #let x = { ... }
+]]
+-- This is very limited, since it would only work correctly if no nested structures (w/ brackets) are found inside
+-- otherwise (if they're found inside), the first closing bracket of that nested structure would close the whole embedded
+-- script, causing the rest of the script to not be treated as a part of embedded script
+local start = (lex:tag(lexer.KEYWORD, P('#') * lex:word_match(lexer.KEYWORD)) *
+	      #((lexer.any - S('{;\n'))^1 * S('{') * lexer.space^0)) +
+	      lex:tag(lexer.OPERATOR,P('#') * S('{'))
+local embed_start = lex:tag('emb_tag', start)
+local embed_end = lexer:tag('emb_tag', S('}'))
```
orbitalquark: I would suggest using `lexer.EMBEDDED` tag names so they are styled correctly.

lexers/typst.lua
```diff
@@ -0,0 +1,125 @@
...
+    -- TODO: Do we really need to not tag a number if procceded by an alpha
+    comment = lex:tag(lexer.COMMENT, lexer.range('/*', '*/') + lexer.to_eol('//')),
+
+    keyword = lex:tag(lexer.KEYWORD, keyword_match),
+
+    header = header(6) + header(5) + header(4) + header(3) + header(2) + header(1)
+  }
+end
+
+local emb_lex = lexer.new('scripting')
```
orbitalquark: I would use a more descriptive name like 'typst_scripting'.

lexers/typst.lua
```diff
@@ -0,0 +1,125 @@
...
+  lexer_obj:add_rule('string', lex:tag(lexer.STRING, rules.string))
+  lexer_obj:add_rule('link', lex:tag(lexer.LINK, rules.link))
+  lexer_obj:add_rule('math', lex:tag('environment.math', rules.math))
+  lexer_obj:add_rule('keyword', rules.keyword)
+  lexer_obj:add_rule('identifier', rules.iden)
+  --lexer_obj:add_rule('number', lex:tag(lexer.NUMBER, rules.numeric_value))
+  lexer_obj:add_rule('list', rules.list)
+  lexer_obj:add_rule('comment', rules.comment)
+  lexer_obj:add_rule('operator', rules.operator)
+end
```
orbitalquark: I feel like we can bring the contents of `build_rules()` into this function. When trying to read this function, I find myself constantly scrolling up to see how each pattern is defined. Making changes would be a bit difficult.

lexers/typst.lua
```diff
@@ -0,0 +1,125 @@
...
+local token = lexer.token
+local P, S, B = lpeg.P, lpeg.S, lpeg.B
+
+local lex = lexer.new(...)
+-- Keep things simple for now and only allow bold and italic in non-code mode
+local italic = -B('\\') * lex:tag(lexer.ITALIC, lexer.range('_', '_'))
+local bold = -B('\\') * lex:tag(lexer.BOLD, lexer.range('*', '*'))
+
+lex:add_rule('bold', bold)
+lex:add_rule('italic', italic)
```
orbitalquark, line 9: Is there a reason to keep these out of `build_rules()`?

lexers/typst.lua
```diff
@@ -0,0 +1,125 @@
...
+end
+
+-- Keywords, functions... don't need '#' when in code
+-- the character `#` is not valid in code
+-- TODO: only enable styling and text related rules when in []
+add_rules(emb_lex, '')
+
+lex:embed(emb_lex, embed_start, embed_end)
+
+add_rules(lex, '#')
```
orbitalquark, line 102: Okay, I'm completely blown away. You've partially solved the "language embedded in itself" problem. I would *really* like to use something like

```
local self = lexer.load('typst', 'typst_scripting')
lex:embed(self, start_rule, end_rule)
```

You cannot do this now because of an endless loop, but I would totally work on a fix to try and make this viable.

lexers/typst.lua
```diff
@@ -0,0 +1,125 @@
...
+
+
+lex:set_word_list(lexer.KEYWORD, {
+  'if', 'else', 'for', 'while', 'let', 'set', 'import', 'include', 'return',
```
orbitalquark, line 106: It's a bit of a nitpick for now, but in my brief testing,

```
#while n < 10 {
  n = (n * 2) - 1
  (n,)
}
```

the `#while` was not highlighting as a keyword.

I just thought I'd make a note.

EDIT: I saw you already noted something similar in an earlier comment, so this is just another instance to report.

orbitalquark (changes requested):
Thanks for your work on this! It looks really promising. As I've noted in a comment below, you've partially solved the "lexer embedded in itself" problem, and I am eager to iterate on it. In the meantime, I've left some comments and suggestions.

This is not a thorough review, so I may have more to say in a subsequent look-over, but it's a start. Thanks again!

EDIT: I realize I haven't addressed any of your inline code comments. Sorry about that. I will do so in a subsequent review. Hopefully there's enough for you to work on until then.

orbitalquark:
Sorry to commit directly to your PR. I was trying to figure out how to show a proof-of-concept for how we can embed this lexer within itself: https://github.com/orbitalquark/scintillua/pull/149/commits/1a3f6f978ba95ad25c2196c77c0c0d26e374b5f7. I thought I'd be on my own branch or something, but I guess not. I'm not very well versed with how git and GitHub work together.

Sorry about all the formatting changes. I forgot to turn off my autoformatter before I started making changes. Hopefully you get the idea.

lexers/typst.lua
ghost: I've tried to keep build_rules generic, so it has the shared rules in both scripting/text mode, as for bold/italic, these must only be applied in text mode, as an example:
```
#let countwords(s) = {
  if s == "" {
    return 0 * 0 * 5
  }
}
```

In the above example if we kept the bold/italic in the build_rules, it would tag the 0 in `0 * 0 * 5` as **bold**, which is not correct, since in scripting mode, -plain- text must be in brackets, and in consequence the italic/bold must be only applied in that case, as an example:
```
#let somefn(word) = {
   let text = [This is a *plain* text];
}
```

ghost:
Thanks so much for the notes, ill try to apply the changes required, as for the embedding mode, i'll look more into it later, and try to find something to fix the unbalanced brackets/curly braces and the recursive tagging in embedding mode

P.S: the [docs](https://orbitalquark.github.io/scintillua/api.html) are very very helpful and straight forward, probably one of the best docs i've came across for a long time since the BSDs' manpages, usually i'd spam some LLM w/ the man page or the docs and ask him what i want (i'm looking at you GNU manpages), but the Scintillua API docs had exactly what i needed

mcepl:
> Thanks for starting on this! When you think it's ready, request a review and I'll have a look. Also, please give a link to Typst documentation.

This is probably what @orbitalquark was after, isn’t it? https://typst.app/docs/reference/syntax/

mcepl:
I wonder whether we shouldn’t have some library of testing documents. I have converted my last sermon slides to Typst (and I am not completely persuaded that it is easier to use than normal LaTeX Beamer, but that isn’t the point here), and I don’t see much in results.

[Ezra 9–10_ Spiritual Reformation.typ.gz](https://github.com/user-attachments/files/20605570/Ezra.9.10_.Spiritual.Reformation.typ.gz)

2ta00ha3:
one of the main limitations in the current lexer impl is the lack of support for recursive lexer embedding,for example, handling transitions between text and embedded scripting languages, the areas where this limitation appears are marked with 'EMBED' in the screenshot:
image

ghost:
one of the main limitations in the current lexer impl is the lack of support for recursive lexer embedding,for example, handling transitions between text and embedded scripting mode, the areas where this limitation appears are marked with 'EMBED' in the screenshot:
image

PR #151, opened 2025-04-01
feat: AsciiDoc lexer

### LINKED ISSUE #104: Add Asciidoc lexer
Here are references from other projects based on Scintilla:

http://hg.code.sf.net/p/scintilla/scite/file/tip/src/asciidoc.properties

https://github.com/notepad-plus-plus/userDefinedLanguages/blob/master/UDLs/AsciiDocWebPage_byEduardoSantana.xml

mcepl:
Based on the top of #138

Fixes: https://github.com/orbitalquark/scintillua/issues/104

### DIFF
--- a/lexers/asciidoc.lua
+++ b/lexers/asciidoc.lua
@@ -1,0 +1,110 @@
+-- Copyright 2025 Matěj Cepl. See LICENSE.
+-- Asciidoc LPeg lexer.
+
+local lexer = lexer
+local P, S, B = lpeg.P, lpeg.S, lpeg.B
+
+local lex = lexer.new(...)
+
+-- Distinguish between horizontal and vertical space so html start rule has a chance to match.
+lex:modify_rule('whitespace', lex:tag(lexer.WHITESPACE, S(' \t')^1 + S('\r\n')^1))
+
+-- Operators.
+lex:add_rule('operator', lex:tag(lexer.OPERATOR, S("*:^'`|~+<=>")))
+
+-- Keywords
+lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD)))
+lex:set_word_list(lexer.KEYWORD, {"NOTE", "IMPORTANT", "WARNING", "TIP", "CAUTION", "TESTME"})
+
+-- Block elements.
+local function h(n)
+	return lex:tag(string.format('%s.h%s', lexer.HEADING, n),
+		lexer.to_eol(lexer.starts_line(string.rep('#', n))))
+end
+lex:add_rule('header', h(6) + h(5) + h(4) + h(3) + h(2) + h(1))
+
+lex:add_rule('hr',
+	lex:tag('hr', lpeg.Cmt(lexer.starts_line(lpeg.C(S('*-_')), true), function(input, index, c)
+		local line = input:match('[^\r\n]*', index):gsub('[ \t]', '')
+		if line:find('[^' .. c .. ']') or #line < 2 then return nil end
+		return (select(2, input:find('\r?\n', index)) or #input) + 1 -- include \n for eolfilled styles
+	end)))
+
+lex:add_rule('list', lex:tag(lexer.LIST,
+	lexer.starts_line(lexer.digit^1 * '.' + S('*+-'), true) * S(' \t')))
+
+local hspace = lexer.space - '\n'
+local blank_line = '\n' * hspace^0 * ('\n' + P(-1))
+
+local code_line = lexer.starts_line((B('    ') + B('\t')) * lexer.to_eol(), true)
+local code_block =
+	lexer.range(lexer.starts_line('```', true), '\n```' * hspace^0 * ('\n' + P(-1))) +
+		lexer.range(lexer.starts_line('~~~', true), '\n~~~' * hspace^0 * ('\n' + P(-1)))
+local code_inline = lpeg.Cmt(lpeg.C(P('`')^1), function(input, index, bt)
+	-- `foo`, ``foo``, ``foo`bar``, `foo``bar` are all allowed.
+	local _, e = input:find('[^`]' .. bt .. '%f[^`]', index)
+	return (e or #input) + 1
+end)
+lex:add_rule('block_code', lex:tag(lexer.CODE, code_line + code_block + code_inline))
+
+lex:add_rule('blockquote',
+	lex:tag(lexer.STRING, lpeg.Cmt(lexer.starts_line('>', true), function(input, index)
+		local _, e = input:find('\n[ \t]*\r?\n', index) -- the next blank line (possibly with indentation)
+		return (e or #input) + 1
+	end)))
+
+-- Span elements.
+lex:add_rule('escape', lex:tag(lexer.DEFAULT, P('\\') * 1))
+
+local link_text = lexer.range('[', ']', true)
+local link_target =
+	'(' * (lexer.any - S(') \t'))^0 * (S(' \t')^1 * lexer.range('"', false, false))^-1 * ')'
+local link_url = 'http' * P('s')^-1 * '://' * (lexer.any - lexer.space)^1 +
+	('<' * lexer.alpha^2 * ':' * (lexer.any - lexer.space - '>')^1 * '>')
+lex:add_rule('link', lex:tag(lexer.LINK, P('!')^-1 * link_text * link_target + link_url))
+
+local link_ref = lex:tag(lexer.REFERENCE, link_text * S(' \t')^0 * lexer.range('[', ']', true))
+local ref_link_label = lex:tag(lexer.REFERENCE, lexer.range('[', ']', true) * ':')
+local ws = lex:get_rule('whitespace')
+local ref_link_url = lex:tag(lexer.LINK, (lexer.any - lexer.space)^1)
+local ref_link_title = lex:tag(lexer.STRING, lexer.range('"', true, false) +
+	lexer.range("'", true, false) + lexer.range('(', ')', true))
+lex:add_rule('link_ref', link_ref + ref_link_label * ws * ref_link_url * (ws * ref_link_title)^-1)
+
+local punct_space = lexer.punct + lexer.space
+
+-- Handles flanking delimiters as described in
+-- https://github.github.com/gfm/#emphasis-and-strong-emphasis in the cases where simple
+-- delimited ranges are not sufficient.
+local function flanked_range(s, not_inword)
+	local fl_char = lexer.any - s - lexer.space
+	local left_fl = B(punct_space - s) * s * #fl_char + s * #(fl_char - lexer.punct)
+	local right_fl = B(lexer.punct) * s * #(punct_space - s) + B(fl_char) * s
+	return left_fl * (lexer.any - blank_line - (not_inword and s * #punct_space or s))^0 * right_fl
+end
+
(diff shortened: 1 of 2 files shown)


lexers/asciidoc.lua
```diff
@@ -0,0 +1,110 @@
...
+lex:set_word_list(lexer.KEYWORD, {"NOTE", "IMPORTANT", "WARNING", "TIP", "CAUTION", "TESTME"})
+
+-- Block elements.
+local function h(n)
+	return lex:tag(string.format('%s.h%s', lexer.HEADING, n),
+		lexer.to_eol(lexer.starts_line(string.rep('#', n))))
+end
+lex:add_rule('header', h(6) + h(5) + h(4) + h(3) + h(2) + h(1))
+
+lex:add_rule('hr',
```
orbitalquark, line 23: This rule will never match for `*`, which is highlighted as an operator.

lexers/asciidoc.lua
```diff
@@ -0,0 +1,110 @@
...
+	lex:tag('hr', lpeg.Cmt(lexer.starts_line(lpeg.C(S('*-_')), true), function(input, index, c)
+		local line = input:match('[^\r\n]*', index):gsub('[ \t]', '')
+		if line:find('[^' .. c .. ']') or #line < 2 then return nil end
+		return (select(2, input:find('\r?\n', index)) or #input) + 1 -- include \n for eolfilled styles
+	end)))
+
+lex:add_rule('list', lex:tag(lexer.LIST,
```
orbitalquark, line 30: This rule will never match for `*`, which is highlighted as an operator.

lexers/asciidoc.lua
```diff
@@ -0,0 +1,110 @@
...
+	lexer.starts_line(lexer.digit^1 * '.' + S('*+-'), true) * S(' \t')))
+
+local hspace = lexer.space - '\n'
+local blank_line = '\n' * hspace^0 * ('\n' + P(-1))
+
+local code_line = lexer.starts_line((B('    ') + B('\t')) * lexer.to_eol(), true)
+local code_block =
```
orbitalquark, line 37: I'm pretty sure these will never match since '`' is highlighted as an operator.

lexers/asciidoc.lua
```diff
@@ -0,0 +1,110 @@
...
+	lexer.range(lexer.starts_line('```', true), '\n```' * hspace^0 * ('\n' + P(-1))) +
+		lexer.range(lexer.starts_line('~~~', true), '\n~~~' * hspace^0 * ('\n' + P(-1)))
+local code_inline = lpeg.Cmt(lpeg.C(P('`')^1), function(input, index, bt)
+	-- `foo`, ``foo``, ``foo`bar``, `foo``bar` are all allowed.
+	local _, e = input:find('[^`]' .. bt .. '%f[^`]', index)
+	return (e or #input) + 1
+end)
+lex:add_rule('block_code', lex:tag(lexer.CODE, code_line + code_block + code_inline))
+
+lex:add_rule('blockquote',
```
orbitalquark, line 47: This rule will not match since `>` is highlighted as an operator.

lexers/asciidoc.lua
```diff
@@ -0,0 +1,110 @@
...
+-- https://github.github.com/gfm/#emphasis-and-strong-emphasis in the cases where simple
+-- delimited ranges are not sufficient.
+local function flanked_range(s, not_inword)
+	local fl_char = lexer.any - s - lexer.space
+	local left_fl = B(punct_space - s) * s * #fl_char + s * #(fl_char - lexer.punct)
+	local right_fl = B(lexer.punct) * s * #(punct_space - s) + B(fl_char) * s
+	return left_fl * (lexer.any - blank_line - (not_inword and s * #punct_space or s))^0 * right_fl
+end
+
+local asterisk_strong = flanked_range('**')
```
orbitalquark, line 83: I don't think this will ever match since `*` is highlighted as an operator.

lexers/asciidoc.lua
```diff
@@ -0,0 +1,110 @@
...
+lex:add_rule('strong', lex:tag(lexer.BOLD, asterisk_strong + underscore_strong))
+
+local asterisk_em = flanked_range('*')
+local underscore_em = (B(punct_space) + #lexer.starts_line('_')) * flanked_range('_', true) *
+	#(punct_space + -1)
+lex:add_rule('em', lex:tag(lexer.ITALIC, asterisk_em + underscore_em))
+
+-- Embedded HTML.
+local html = lexer.load('html')
+local start_rule = lexer.starts_line(P(' ')^-3) * #P('<') * html:get_rule('tag') -- P(' ')^4 starts code_line
```
orbitalquark, line 95: I'm pretty sure this will never match since `<` is highlighted as an operator.

orbitalquark (changes requested):
This is mostly a copy-paste of the Markdown lexer with 3 new rules. The operator rule is most problematic because it matches characters that subsequent rules depend on, invalidating much of the lexer.

If asciidoc is a superset of Markdown, then please inherit from Markdown and use `lex:modify_rule()` and `lex:add_rule()` to tag new elements.

lexers/asciidoc.lua
mcepl: You are right, this is mostly based on something I found somewhere on the Web. And yes, you are also right that `operator` rule has to go. What does operator actually mean in the context of a document? Will remove.

Or … would it be possible somehow limit a rule to some context? I.e., make this rule work only inside the context of mathematical expressions ([`stem`](https://docs.asciidoctor.org/asciidoc/latest/stem/))? However, current lexer doesn’t support `stem` anyway, so this is moot.

lexers/asciidoc.lua
mcepl: see above

lexers/asciidoc.lua
mcepl: see above

lexers/asciidoc.lua
mcepl: ditto

lexers/asciidoc.lua
mcepl: see above

lexers/asciidoc.lua
mcepl: see above

lexers/asciidoc.lua
orbitalquark: Rules can have limited backtracking, like `lex:add_rule('method', -lpeg.B('.') * lex:tag(lexer.METHOD, lexer.word) * #P('('))`, but it's often not possible with more complicated contexts because the lexer can lose track of state. You can try making use of the `lexer.line_state` table, but it's hard to get it right.

orbitalquark (changes requested):
My original review still stands: this looks more-or-less like a copy-paste of the Markdown lexer into asciidoc. If asciidoc truly inherits from markdown, then please use proper inheritance.

There is a maintenance burden with lexers that are sub-optimally written. The markdown lexer is going to get a refactor one of these days. The asciidoc lexer would have to have a similar refactor if it didn't already inherit from markdown.

mcepl:
> My original review still stands: this looks more-or-less like a copy-paste of the Markdown lexer into asciidoc. If asciidoc truly inherits from markdown, then please use proper inheritance.
> 
> There is a maintenance burden with lexers that are sub-optimally written. The markdown lexer is going to get a refactor one of these days. The asciidoc lexer would have to have a similar refactor if it didn't already inherit from markdown.

I don’t dispute you much, I just wonder how much it makes sense. Markdown is a silly little hack which got completely out of hand and now has a zillion different flavours ([Common Mark](https://commonmark.org) and [GitHub Flavour](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github) are two most used ones). [AsciiDoc](https://docs.asciidoctor.org/asciidoc/latest/) is its almost complete opposite: originating in super-complicated DocBook XML, coming from (among other places) O’Reilly publishing house, having a full-blown standard available. Yes, although all three (or four) major lightweight markup languages (together with reStrcturedText and perhaps Wikipedia text) are kind of similar to each other, [there are plenty of differences](https://docs.asciidoctor.org/asciidoc/latest/asciidoc-vs-markdown/), and I think maintaining AsciiDoc on the top of Markdown could get rather complicated quite quickly.

Anyway, I can certainly rewrite the thing any time.

orbitalquark:
Please understand that I know nothing about AsciiDoc. I only noticed your PR included rules that looked similar to Markdown's rules. Upon closer inspection, I noticed the copy-paste nature and pointed it out. If AsciiDoc is dissimilar to Markdown, then I agree inheritance is out of the question, but the rules will need to be rewritten.

Thanks for sharing your thoughts.

mcepl:
On Wed Apr 9, 2025 at 3:00 PM CEST, orbitalquark wrote:
> Please understand that I know nothing about AsciiDoc. I only noticed your PR included rules that looked similar to Markdown's rules. Upon closer inspection, I noticed the copy-paste nature and pointed it out. If AsciiDoc is dissimilar to Markdown, then I agree inheritance is out of the question, but the rules will need to be rewritten.

AsciiDoc is certainly not a derivative of Markdown. Long time ago, beyond seven mountains and seven rivers, there were times when people were actually reading books printed with black ink on white paper. In those times, one of the best publishers, or a manufacturer, of such books related to computers, including celebrated The Lama, The Rhino, and The Bat Books, was O’Reilly & Associates. They decided to publish books authored in DocBook XML, a very baroque dialect of XML, describing very semantically and correctly everything in a book down to `<warning>`, `<variable>`, and `<footnote>`. Did you ever write a large book in DocBook XML? The result is certainly awesome, but the process of writing is nothing very funny or something what anybody who experienced it would love to repeat it. So, when Markdown (and rST) happened, all their authors wanted to use that instead. O’Reilly didn’t allow that, because they wanted to preserve richness of DocBook, but they developed their own lightweight markdown language corresponding to DocBook, which is what AsciiDoc is.

orbitalquark (changes requested):
I'm sorry to keep insisting, but this lexer is pretty much a copy-paste of the markdown lexer. You are saying that ASCIIDoc is not a derivative of markdown, yet the code you are submitting is nearly an exact copy of the markdown lexer.

I cannot accept this discrepancy. Please either change this lexer to inherit from markdown, or rewrite the lexer to support ASCIIDoc, which is not markdown.

mcepl:
OK, let’s kill it here. I don’t have enough motivation to rewrite it from ground, because I don’t care for AsciiDoc enough. If anybody cares, they can do the work.

M0JXD:
Hey all, I'm interested in getting this over the finish line either as standalone or inherited. But before doing that I'm not sure this lexer was ever accurate? E.g. it's using the Mardown lexer's `#` rule for headers but Asciidoc uses `=`, and various other things are not correct like indented lists and bold text.

[Asciidoc](https://asciidoc.org/) shows a comparison to Markdown on their main page. In their own words:

> AsciiDoc appears strikingly similar to Markdown, making way for an easy transition. Where AsciiDoc shines is in its depth. AsciiDoc provides all the essential elements in technical writing out of the box. No variants needed. Its syntax can be elaborated without having to fundamentally change the language, assuring users that it’s still standard AsciiDoc.

I think an accurate Asciidoc lexer would end up quite different to the Markdown one, but I don't want to start work without a consensus as to what is best. I'm happy to make an effort to inherit from Markdown if that would still be preferred.

PR #154, opened 2025-04-22
fix(python): Python introduced new t-strings

mcepl:
References: https://peps.python.org/pep-0750
References: https://davepeck.org/2025/04/11/pythons-new-t-strings

### DIFF
--- a/lexers/python.lua
+++ b/lexers/python.lua
@@ -33,7 +33,7 @@ lex:add_rule('constant', builtin_const + attr)
 local sq_str = lexer.range("'", true)
 local dq_str = lexer.range('"', true)
 local tq_str = lexer.range("'''") + lexer.range('"""')
-lex:add_rule('string', lex:tag(lexer.STRING, (S('fFrRbBrR') * S('rRfFrRbB') + S('ruRUfFbB'))^-1 *
+lex:add_rule('string', lex:tag(lexer.STRING, (S('fFrRbBrRtT') * S('rRfFrRbBtT') + S('ruRUfFbBtT'))^-1 *
 	(tq_str + sq_str + dq_str)))
 
 -- Identifiers.


mcepl:
What’s missing here?

lexers/python.lua
```diff
@@ -33,7 +33,7 @@ lex:add_rule('constant', builtin_const + attr)
 local sq_str = lexer.range("'", true)
 local dq_str = lexer.range('"', true)
 local tq_str = lexer.range("'''") + lexer.range('"""')
-lex:add_rule('string', lex:tag(lexer.STRING, (S('fFrRbBrR') * S('rRfFrRbB') + S('ruRUfFbB'))^-1 *
+lex:add_rule('string', lex:tag(lexer.STRING, (S('fFrRbBrRtT') * S('rRfFrRbBtT') + S('ruRUfFbBtT'))^-1 *
```
orbitalquark, line 36: I honestly don't understand why this prefix pattern is written this way. I'm looking through the revision history and it looks like I added it here: https://github.com/orbitalquark/scintillua/commit/051d2ec00859bcf48c32f653309140dbf0e43e99#diff-46651a8c375626d6c358fc27f41c182bf025b6b95bb9435e2187fe6ab77775deR35, but I don't understand why I did that 😕 

This needs to be cleaned up before we can add the `t` prefix. Also, is a `T` prefix allowed? I couldn't find a reference to it in the PEP.

orbitalquark:
I cleaned up the prefix literals and added `tT` in https://github.com/orbitalquark/scintillua/commit/c481059400db8d31b0d025f869e8aec94c08d36c.

lexers/python.lua
mcepl: I know I am late, but this is the canonical documentation: https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals .

orbitalquark: Thank you! I was looking for this but couldn't find it. Committed a new change: https://github.com/orbitalquark/scintillua/commit/7411179392356af180b0f433b8064248116a6d18.

PR #156, opened 2025-05-28
fix: modernize ini lexer

mcepl:
Not completely certain about

```lua
local sq_str = P('L')^-1 * lexer.range("'")
local dq_str = P('L')^-1 * lexer.range('"')
```

Yes, Lua is important for us, but `L` prefix is rather non-standard in the `ini` universe, isn’t it?

### DIFF
--- a/lexers/ini.lua
+++ b/lexers/ini.lua
@@ -1,38 +1,39 @@
 -- Copyright 2006-2025 Mitchell. See LICENSE.
 -- Ini LPeg lexer.
 
-local lexer = require('lexer')
-local token, word_match = lexer.token, lexer.word_match
+local lexer = lexer
 local P, S = lpeg.P, lpeg.S
 
-local lex = lexer.new('ini')
-
--- Whitespace.
-lex:add_rule('whitespace', token(lexer.WHITESPACE, lexer.space^1))
+local lex = lexer.new(...)
 
 -- Keywords.
-lex:add_rule('keyword', token(lexer.KEYWORD, word_match('true false on off yes no')))
+lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD)))
 
 -- Identifiers.
-lex:add_rule('identifier', token(lexer.IDENTIFIER, (lexer.alpha + '_') * (lexer.alnum + S('_.'))^0))
+-- Allows periods within identifiers for dotted keys (e.g., 'foo.bar').
+lex:add_rule('identifier',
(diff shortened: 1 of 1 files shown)


lexers/ini.lua
```diff
@@ -1,38 +1,39 @@
...
+-- Allows periods within identifiers for dotted keys (e.g., 'foo.bar').
+lex:add_rule('identifier',
+    lex:tag(lexer.IDENTIFIER, (lexer.alpha + '_') * (lexer.alnum + S('_. '))^0))
 
 -- Strings.
-local sq_str = lexer.range("'")
-local dq_str = lexer.range('"')
-lex:add_rule('string', token(lexer.STRING, sq_str + dq_str))
+local sq_str = P('L')^-1 * lexer.range("'")
+local dq_str = P('L')^-1 * lexer.range('"')
```
orbitalquark: I don't think there's a need for this prefix. I haven't seen this in any ini file before. Have you?

lexers/ini.lua
```diff
@@ -1,38 +1,39 @@
...
+	lex:tag(lexer.COMMENT, lexer.to_eol(lexer.starts_line(S(';#')))))
 
 -- Numbers.
-local integer = S('+-')^-1 * (lexer.hex_num + lexer.oct_num_('_') + lexer.dec_num_('_'))
-lex:add_rule('number', token(lexer.NUMBER, lexer.float + integer))
+lex:add_rule('number', lex:tag(lexer.NUMBER, lexer.float + lexer.integer))
 
 -- Operators.
-lex:add_rule('operator', token(lexer.OPERATOR, '='))
+lex:add_rule('operator', lex:tag(lexer.OPERATOR, S('=:')))
```
orbitalquark, line 33: May I ask where you see a need for ':' as an operator? I've seen Windows paths (e.g. c:\foo) used in ini files and don't think the ':' should be highlighted as an operator.

orbitalquark (changes requested):
Thanks for submitting this. I have a couple of questions.

lexers/ini.lua
mcepl: I thought it is something Lua-related, but it isn’t. Removed.

lexers/ini.lua
mcepl: Many `ini`-like files use `:` instead of `=` to separate name of the variables from its value. E.g., this could be valid config file for the Python `configparser` library:

```ini
[general]
name: value
```

Of course, it works poorly with Windows path names, but not everybody is that much concerned with those.

orbitalquark:
Thanks for the feedback. Committed in https://github.com/orbitalquark/scintillua/commit/7c82a90afe6b0e4841163f33721aa4252c441fab.

PR #158, opened 2025-05-28
fix: modernize zig lexer

mcepl:

### DIFF
--- a/lexers/zig.lua
+++ b/lexers/zig.lua
@@ -20,10 +20,18 @@ lex:add_rule('constant', lex:tag(lexer.CONSTANT, lex:word_match(lexer.CONSTANT))
 lex:add_rule('function',
 	lex:tag(lexer.FUNCTION_BUILTIN, '@' * lex:word_match(lexer.FUNCTION_BUILTIN)))
 
+-- Bare functions (function calls without @).
+lex:add_rule('function_call',
+	lex:tag(lexer.FUNCTION, lexer.word * #(lexer.space^0 * '(')))
+
+-- Struct methods (word followed by dot and another word).
+lex:add_rule('method', lex:tag(lexer.FUNCTION, lexer.word * '.' * lexer.word))
+
 -- Strings.
+local raw_str = P('\\\\') * lexer.range('\n', false, false) -- Multi-line strings
 local sq_str = lexer.range("'", true) -- Character/Byte literal
 local dq_str = lexer.range('"', true) -- String/Byte array literal
-lex:add_rule('string', lex:tag(lexer.STRING, sq_str + dq_str))
+lex:add_rule('string', lex:tag(lexer.STRING, raw_str + sq_str + dq_str))
 
 -- Identifiers.
 lex:add_rule('identifier', lex:tag(lexer.IDENTIFIER, lexer.word))
@@ -34,15 +42,20 @@ local comment = lexer.to_eol('//', true)     -- Single-line comments
 lex:add_rule('comment', lex:tag(lexer.COMMENT, doc_comment + comment))
 
 -- Numbers.
-lex:add_rule('number', lex:tag(lexer.NUMBER, lexer.number))
+local binary = P('0b') * S('01')^1
+local octal = P('0o') * S('01234567')^1  
+local hex = P('0x') * S('0123456789abcdefABCDEF')^1
+local decimal = lexer.number
+lex:add_rule('number', lex:tag(lexer.NUMBER, binary + octal + hex + decimal))
 
 -- Operators.
-lex:add_rule('operator', lex:tag(lexer.OPERATOR, S('+-/*%<>!=^&|?~:;,.()[]{}')))
+lex:add_rule('operator', lex:tag(lexer.OPERATOR, S('+-/*%<>!=^&|?~:;,.()[]{}') +
+	P('**') + P('<<') + P('>>') + P('==') + P('!=') + P('<=') + P('>=')))
 
 -- Word lists
 lex:set_word_list(lexer.KEYWORD, {
 	-- Keywords.
-	'inline', 'pub', 'fn', 'comptime', 'const', 'extern', 'return', 'var', 'usingnamespace',
+	'inline', 'pub', 'fn', 'comptime', 'const', 'return', 'var', 'usingnamespace',
 	-- Defering code blocks.
 	'defer', 'errdefer',
 	-- Functions and structures related keywords.
@@ -50,7 +63,7 @@ lex:set_word_list(lexer.KEYWORD, {
 	'callconv', 'packed', 'linksection', 'unreachable', 'test', 'asm',
 	'volatile',
 	-- Parallelism and concurrency related keywords.
-	'async', 'await', 'noasync', 'suspend', 'nosuspend', 'resume', 'threadlocalanyframe',
+	'async', 'await', 'noasync', 'suspend', 'nosuspend', 'resume', 'threadlocal', 'anyframe',
 	-- Control flow: conditions and loops.
 	'if', 'else', 'orelse', 'or', 'and', 'while', 'for', 'switch', 'continue', 'break', 'catch',
 	'try',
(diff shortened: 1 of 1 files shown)


lexers/zig.lua
```diff
@@ -27,29 +54,22 @@ lex:add_rule('keyword', token(lexer.KEYWORD, word_match{
...
+	'c_void', -- C void type
+	'f16', 'f32', 'f64', 'f128', -- Floating-point types
+	'bool', 'void', 'noreturn', 'type', 'anytype', 'error', 'anyerror', -- Special types
+	'comptime_int', 'comptime_float' -- Comptime types
+})
 
--- Built-in functions.
-lex:add_rule('function', token(lexer.FUNCTION, '@' * word_match{
+lex:set_word_list(lexer.FUNCTION, {
+	-- Extensive list of @-prefixed built-in functions
```
orbitalquark: According to https://ziglang.org/documentation/master/#Builtin-Functions, there are more functions like `addrSpaceCast` and `branchHint`, while some functions like `asyncCall` and `boolToInt` no longer exist. This list should be updated.

lexers/zig.lua
```diff
@@ -2,17 +2,44 @@
...
+lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD)))
+
+-- Types.
+lex:add_rule('type', lex:tag(lexer.TYPE, lex:word_match(lexer.TYPE)))
+
+-- Constants.
+lex:add_rule('constant', lex:tag(lexer.CONSTANT, lex:word_match(lexer.CONSTANT)))
+
+-- Built-in functions.
+lex:add_rule('function', lex:tag(lexer.FUNCTION, '@' * lex:word_match(lexer.FUNCTION)))
```
orbitalquark: These ought to have the `_BUILTIN` suffix.

orbitalquark (changes requested):
Thanks for starting the work on this. Migrating lexers is an opportunity to revisit what may be missing from the lexer.

- It looks to me like Zig has struct methods (https://ziglang.org/documentation/master/#struct), so we should support them if possible, like in the C lexer.
- Can we highlight bare functions, like in the C lexer? (i.e. `lexer.word * #(lexer.space^0 * '('`)
- There may be more things to consider that I haven't thought of.

lexers/zig.lua
mcepl: I added all missing functions, but I have not removed anything, because I guess, even obsolete Zig should be parsed well, shouldn’t they? Or do we want to try to be Zig linter as well?

lexers/zig.lua
mcepl: You mean 

```lua
lex:add_rule('function', lex:tag(lexer.FUNCTION_BUILTIN, '@' * lex:word_match(lexer.FUNCTION_BUILTIN)))
```

lexers/zig.lua
orbitalquark: I personally would remove them since Zig is not 1.0 and in heavy development; I doubt any serious users would be using an old version.

mcepl:
Just FYI, there is a complete grammar as PEG, https://ziglang.org/documentation/master/#Grammar

lexers/zig.lua
```diff
@@ -2,88 +2,114 @@
...
-	'comptime_int', 'comptime_float'
-}))
+lex:add_rule('type', lex:tag(lexer.TYPE, lex:word_match(lexer.TYPE)))
 
 -- Constants.
-lex:add_rule('constant', token(lexer.CONSTANT, word_match{
-	-- Special values.
-	'false', 'true', 'null', 'undefined'
-}))
+lex:add_rule('constant', lex:tag(lexer.CONSTANT, lex:word_match(lexer.CONSTANT)))
```
orbitalquark: This should be `CONSTANT_BUILTIN`.

lexers/zig.lua
```diff
@@ -2,88 +2,114 @@
...
-}))
+lex:add_rule('function',
+	lex:tag(lexer.FUNCTION_BUILTIN, '@' * lex:word_match(lexer.FUNCTION_BUILTIN)))
+
+-- Bare functions (function calls without @).
+lex:add_rule('function_call',
+	lex:tag(lexer.FUNCTION, lexer.word * #(lexer.space^0 * '(')))
+
+-- Struct methods (word followed by dot and another word).
+lex:add_rule('method', lex:tag(lexer.FUNCTION, lexer.word * '.' * lexer.word))
```
orbitalquark: Given `const math = std.math;`, this will incorrectly tag `std.math` as a function.

Also, your comment says these are struct methods, but I don't think `std.math.inf()` is a struct method. Also, instead of tagging with `lexer.FUNCTION`, methods are tagged with `lexer.FUNCTION_METHOD`.

lexers/zig.lua
```diff
@@ -2,88 +2,114 @@
...
 
 -- Strings.
-local sq_str = P('L')^-1 * lexer.range("'", true)
-local dq_str = P('L')^-1 * lexer.range('"', true)
-lex:add_rule('string', token(lexer.STRING, sq_str + dq_str))
+local raw_str = P('\\\\') * lexer.range('\n', false, false) -- Multi-line strings
```
orbitalquark: Given
```
const hello_world_in_c =
    \\#include <stdio.h>
    \\
    \\int main(int argc, char **argv) {
    \\    printf("hello world\n");
    \\    return 0;
    \\}
;
```
this will not highlight that multi-line string properly.

I suggest `lexer.to_eol('\\\\')`

lexers/zig.lua
```diff
@@ -2,88 +2,114 @@
...
+local comment = lexer.to_eol('//', true)     -- Single-line comments
+lex:add_rule('comment', lex:tag(lexer.COMMENT, doc_comment + comment))
 
 -- Numbers.
-lex:add_rule('number', token(lexer.NUMBER, lexer.number))
+local binary = P('0b') * S('01')^1
+local octal = P('0o') * S('01234567')^1  
+local hex = P('0x') * S('0123456789abcdefABCDEF')^1
+local decimal = lexer.number
+lex:add_rule('number', lex:tag(lexer.NUMBER, binary + octal + hex + decimal))
```
orbitalquark: Apparently zig allows for '_' separator characters to make numbers more readable (https://ziglang.org/documentation/master/#Integers). This can be reduced to `lex:tag(lexer.NUMBER, lexer.number_('_'))`

lexers/zig.lua
```diff
@@ -2,88 +2,114 @@
...
 
 -- Operators.
-lex:add_rule('operator', token(lexer.OPERATOR, '..' + S('+-/*%<>!=^&|?~:;,.()[]{}')))
+lex:add_rule('operator', lex:tag(lexer.OPERATOR, S('+-/*%<>!=^&|?~:;,.()[]{}') +
+	P('**') + P('<<') + P('>>') + P('==') + P('!=') + P('<=') + P('>=')))
```
orbitalquark: None of these will ever be matched. The first `S()` pattern will always match these. I honestly don't see a need for these two-character strings.

Your change also removes support for Zig's range operator, `..`, so `foo[0..1]` is no longer highlighted properly.

orbitalquark (commented):
Let's stop here. I don't think this is fun for either of us to try and push and pull this over the finish line.

There's still a lot of work to be done. I went through maybe the first 1/12th of the Zig master documentation and found unimplemented things like '_'s in integers, multi-line strings that do not work in this implementation, thoughts on if we should support arbitrary 'iN' and 'uN' types where N is an integer like 'i7' or 'u7', and so on.

I would like for both of us to want to give our best effort at producing a modern Zig lexer, not a lexer that checks the minimum amount of boxes beyond a simple migration. This is what I meant in my initial review comment of "Migrating lexers is an opportunity to revisit what may be missing from the lexer."

I'm sorry. This is probably not what you wanted to hear.

mcepl:
> Let's stop here. I don't think this is fun for either of us to try and push and pull this over the finish line.

No, it isn’t. And I think we are hitting differences in opinions. Certainly, it is your project, so you can manage it however you think is the best, but I am much more friend of “release early, release often”, “Worse is better”, the tendency to have at least any support for various document types even though they are not perfect. Volunteer work is based on quick feedback in my opinion and frequent release of dopamine from having the PRs accepted, having PRs opened since March 2024 doesn’t make anybody excited. Also, I believe that you attract more contributors when they have something to contribute on: starting completely new lexer (although we know it is not THAT complicated) seems so daunting than fewer people would do it, rather than adding correct dealing with `struct` types in the already existing Zig lexer, which could be done by people actually using the lexer and who have interest in making it better.

However, as I have already said, it is your project and you can do whatever you want to do with it.

Concerning these PRs of mine: I will gladly fix whatever specific issues you have with them, but diving through many pages long standards is exactly something I don’t want to get involved in.

lexers/zig.lua
mcepl: Reverted.

PR #159, opened 2025-05-29
fix: modernize fantom lexer

mcepl:

### DIFF
--- a/lexers/fantom.lua
+++ b/lexers/fantom.lua
@@ -2,26 +2,52 @@
 -- Fantom LPeg lexer.
 -- Based on Java LPeg lexer by Mitchell and Vim's Fantom syntax.
 
-local lexer = require('lexer')
-local token, word_match = lexer.token, lexer.word_match
+local lexer = lexer
 local P, S = lpeg.P, lpeg.S
 
-local lex = lexer.new('fantom')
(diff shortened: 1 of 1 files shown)


orbitalquark (changes requested):
Thanks for starting work on this. As mentioned in another review, migrating lexers is an opportunity to revisit what may be missing from lexers.

Since Fantom seems to be a Java-like language, we should probably support function and method highlighting. It would also be nice to recognize some commonly used builtin functions/classes/types (https://fantom.org/doc/) like the Java lexer does.

mcepl:
> Since Fantom seems to be a Java-like language, we should probably support function and method highlighting. It would also be nice to recognize some commonly used builtin functions/classes/types (https://fantom.org/doc/) like the Java lexer does.

I have to say zig is somewhere where I draw the line. At least there I have, unfortunately, don’t let me rant on what I think about these modern compiled languages, heard about projects written in it. I have zero-point-zero interest in Fantom, I don’t want to study it just to help its support in Scintillua. Anybody, feel free to push more commits to this PR, if @orbitalquark doesn’t accept it as it is.

mcepl:
No reactions from the maintainer and no interest from me.

PR #160, opened 2025-05-31
rust lexer: remove self as keyword

ludverse:
In Rust, although "self" is technically a keyword, it's used as a regular variable. So to me, when in my code editor [vis](https://github.com/martanne/vis), "self" having the same colour as regular identifiers makes way more sense and makes it more readable.

Lmk what you think. All of this is very subjective.


orbitalquark:
If it's a keyword, then it ought to be recognized as such by default.

If your editor allows access to the lexer, you can call `lex:set_word_list(lexer.KEYWORD, {...})` with your own, `self`-less list. These types of situations are what I had in mind when adding the feature some time ago. Not all lexers support it yet, but Rust does :)

ludverse:
Yup that's fair enough. 

> If your editor allows access to the lexer, you can call lex:set_word_list(lexer.KEYWORD, {...}) with your own, self-less list.

Nice, thanks! And thank you for maintaining this :)

PR #162, opened 2025-06-15
Add Janet Lexer

engenforge:
My initial attempt at a lexer for [Janet](https://github.com/janet-lang/janet).  I believe this addresses most or all of the feedback provided in a [previous pull request](https://github.com/orbitalquark/scintillua/pull/66).

### DIFF
--- a/lexers/janet.lua
+++ b/lexers/janet.lua
@@ -2,7 +2,7 @@
 -- Janet LPeg lexer.
 
 local lexer = lexer
-local P, S = lpeg.P, lpeg.S
+local P, S, B , utfR = lpeg.P, lpeg.S, lpeg.B, lpeg.utfR
 
 local lex = lexer.new(...)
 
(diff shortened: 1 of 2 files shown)


lexers/janet.lua
```diff
@@ -0,0 +1,172 @@
...
+lex:add_rule('comment', lex:tag(lexer.COMMENT, lexer.to_eol('#')))
+
+local sq_str = lexer.range('"') + lexer.range('`')
+local dq_str = lexer.range('``')
+local tq_str = lexer.range('```')
+lex:add_rule('string', lex:tag(lexer.STRING, tq_str + dq_str + sq_str))
+
+lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD)))
+
+lex:add_rule('function', lex:tag(lexer.FUNCTION, lex:word_match(lexer.FUNCTION)))
```
orbitalquark: Let's use `lexer.FUNCTION_BUILTIN` here. You'll also need to change the word list definition on line 49.

lexers/janet.lua
```diff
@@ -0,0 +1,172 @@
...
+
+lex:add_rule('number', lex:tag(lexer.NUMBER, S('-+')^-1 * lexer.digit^1 * (S('._') + lexer.alnum)^0))
+
+local id_ch = S('!@$%^&*-_+=:<>.?') + lexer.alnum
+lex:add_rule('constant', lex:tag(lexer.CONSTANT, P(':') * id_ch^0))
```
orbitalquark, line 46: Is a bare ':' allowed to be a constant? If not, then you probably want to use `id_ch^1`.

lexers/janet.lua
```diff
@@ -0,0 +1,172 @@
...
+
+lex:add_rule('identifier', lex:tag(lexer.IDENTIFIER, id_ch^1))
+
+lex:add_fold_point(lexer.OPERATOR, '(', ')')
```
orbitalquark, line 75: I don't see an operator rule that matches punctuation including parentheses, brackets, and braces. Perhaps you forgot it?

orbitalquark (changes requested):
Thanks for working on this. I've left some feedback. In addition to those items, would you please either update the `lexer.detect()` function to include Julia extensions (https://github.com/orbitalquark/scintillua/blob/7c82a90afe6b0e4841163f33721aa4252c441fab/lexers/lexer.lua#L1676), or let me know what extension to use and I can make the change.

Before committing this, I will likely re-arrange the rules to match other lexers. Typically, keywords, functions, and identifiers come before comments and strings, for example. Feel free to get a head start on that, but I won't require it.

lexers/janet.lua
engenforge: Yes, I was a bit surprised by that as well, but it is allowed per the Janet [documentation](https://github.com/janet-lang/janet-lang.org/blob/eec6780d94e84f2269eb8c22e0ae9d820feba461/content/docs/syntax.mdz#L73).

lexers/janet.lua
engenforge: Good suggestion.  I updated in commit https://github.com/orbitalquark/scintillua/pull/162/commits/33949bc198f3c1dfb7d13f073da1055a62f8651b.

lexers/janet.lua
engenforge: Thank you.  Good catch.  I added it in commit https://github.com/orbitalquark/scintillua/pull/162/commits/56359baace99c194ca514fa1ddc42f66c705222e

engenforge:
@orbitalquark, I also updated the lexer detect function and reordered the rules to match other lexers.  Thanks!

lexers/janet.lua
```diff
@@ -0,0 +1,173 @@
...
+-- reserved by the janet compiler such as nil, true, do, fn, etc.
+
+lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD)))
+
+lex:add_rule('function', lex:tag(lexer.FUNCTION_BUILTIN, lex:word_match(lexer.FUNCTION_BUILTIN)))
+
+local id_ch = S('!@$%^&*-_+=:<>.?') + lexer.alnum
+lex:add_rule('constant', lex:tag(lexer.CONSTANT, P(':') * id_ch^0))
+
+lex:add_rule('identifier', lex:tag(lexer.IDENTIFIER, id_ch^1))
```
orbitalquark: Sorry I missed this in my first review, but this will actually match numbers, so the 'number' rule below will never match.

You could move the number rule up before this one, but then it becomes a question of how Julia distinguishes between numbers and identifiers (if it even does).

lexers/janet.lua
```diff
@@ -0,0 +1,173 @@
...
+local sq_str = lexer.range('"') + lexer.range('`')
+local dq_str = lexer.range('``')
+local tq_str = lexer.range('```')
+lex:add_rule('string', lex:tag(lexer.STRING, tq_str + dq_str + sq_str))
+
+lex:add_rule('comment', lex:tag(lexer.COMMENT, lexer.to_eol('#')))
+
+lex:add_rule('number', lex:tag(lexer.NUMBER, S('-+')^-1 * lexer.digit^1 * (S('._') + lexer.alnum)^0))
+
+lex:add_rule('operator', lex:tag(lexer.OPERATOR, S('<>=*/+-%()[]{}')))
```
orbitalquark: Any of the symbols here that are also defined in `id_ch` will never match as operators; they will match as identifiers instead. Sorry I don't have the time to investigate how Julia actually defines identifiers, but we'll need to figure that out.

orbitalquark (changes requested):
Almost there. There are a couple of tricky things we'll need to resolve first.

lexers/janet.lua
engenforge: Yes, I recall now that I selected my initial order of rules such that numbers come before identifiers for this reason.  I forgot this and blindly reordered per your request without thinking it through thoroughly.  Is it bad practice in Scintillua to have rules that are dependent on order?  The definition in the Janet language is "a sequence of alphanumeric characters not starting with a digit or a colon", so I will update the identifier rule such that the first character is not a digit or colon.  I will double check this for all rules so that they aren't dependent on the order they are added in the lexer.

lexers/janet.lua
engenforge: Same comment as above.  I will fix this so that it isn't dependent on the order the rule is added.

lexers/janet.lua
engenforge: By the way @orbitalquark , I noticed you mentioned Julia multiple times.  I just wanted to confirm that this lexer is for the Janet language to avoid any potential confusion.  Thanks!

orbitalquark: Sorry, I get the two of them confused all the time.

orbitalquark: And no, it is not bad practice to have rules dependent on order. Scintillua matches rules in the order they are defined, so order matters.

lexers/janet.lua
engenforge: I worked on rewriting the rules so that they weren't dependent on order, however it became IMHO more complex than it needed to be.  So in the interest of keeping things simpler I reverted to the earlier order and things seem to be working again.

lexers/janet.lua
engenforge: See [comment below](https://github.com/orbitalquark/scintillua/pull/162#discussion_r2155081962) for my latest thoughts and resolving of this.

lexers/janet.lua
```diff
@@ -0,0 +1,177 @@
...
+-- operator, constant, and identifier need to occur in that order for lexing
+-- rules to function properly.
+
+lex:add_rule('function', lex:tag(lexer.FUNCTION_BUILTIN, lex:word_match(lexer.FUNCTION_BUILTIN)))
+
+lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD)))
+
+lex:add_rule('number', lex:tag(lexer.NUMBER, S('-+')^-1 * lexer.digit^1 * (S('._') + lexer.alnum)^0))
+
+lex:add_rule('operator', lex:tag(lexer.OPERATOR, S("<>=*/+-%&',;~@()[]{}")))
```
orbitalquark: I'm wondering if we should break this up into `S("/',;~()[]{}") + S('<>=*+-%&@') * -lexer.alnum` in order to avoid tagging the first character of an identifier as an operator. What do you think?

engenforge: @orbitalquark, good catch.  I reviewed the lexer closer and spotted a couple other similar issues where rules weren't catching all the edge cases for identifiers, keywords, strings, etc.  I subsequently spent a fair amount of time reworking the lexing rules as part of commit https://github.com/orbitalquark/scintillua/pull/162/commits/3bc75e4bd491952b322b465374bb1de9d7253304.  The string lexer became a little verbose / redundant, but I couldn't figure out an elegant way to simplify that.  Maybe you have a suggestion or advice on that.

engenforge:
On another note I created a Janet source file with a variety of syntax test cases so that I could confirm the lexer was working properly.  I've pasted it below in case it's helpful for anyone else in the future:
```janet
# Janet syntax lexing test file.

# This file is intended to be used to test syntax highlighting of janet source
# code with scintillua. This file can also be successfully parsed by the Janet
# interpreter although the code example at the beginning is the only piece
# that will generate output.

# Below is the code example from https://janet-lang.org
(defn sum3
  "Solve the 3SUM problem in O(n^2) time."
  [s]
  (def tab @{})
  (def solutions @{})
  (def len (length s))
  (for k 0 len
    (put tab (s k) k))
  (for i 0 len
    (for j 0 len
      (def k (get tab (- 0 (s i) (s j))))
      (when (and k (not= k i) (not= k j) (not= i j))
        (put solutions {i true j true k true} true))))
  (map keys (keys solutions)))

(let [arr @[2 4 1 3 8 7 -3 -1 12 -5 -8]]
  (printf "3sum of %j: " arr)
  (printf "%j" (sum3 arr)))

# The following are examples intended to cover a number of edge cases for some
# of the syntax categories.  A number of these examples were obtained from the
# following web page:
# https://janet-lang.org/docs/syntax.html

# Strings (immutable)
(def janet_string_examples [
	"This is a \" ````` string"
	`This is a string \`
	`This is a " string `
	``This is a string \``
	```This is a `` " string```
	````This is a ``` " string````
	````This is
a ``` "
multiline string```` ])

# Strings (mutable. Janet calls these buffers)
(def janet_buffer_examples [
	@""
	@"Buffer."
	@``Another buffer``
	@`
Yet another buffer
`])

# Constants (Janet calls these keywords)
(def janet_constant_examples_from_website
	[ :keyword :range :0x0x0x0 :a-keyword :: : ])

# Identifiers (Janet calls these symbols)
(let [ set😀 1 £😀 1 set+ 1 x 1 +set 1 mset 1 sum3 1 @num 1])
(let [  kebab-case-symbol 1
	snake_case_symbol 1
	my-module/my-function 1
	***** 1
	!%$^*__--__._+++===-crazy-symbol 1
	*global-var* 1
	你好 1])

# Numbers
(def janet_number_examples_from_website
	[ 0 12 -65912 4.98 1.3e18 1.3E18 18r123C 11raaa&a 1_000_000 0xbeef ])

# Tuples
(def tuple1 (do 1 2 3))
(def tuple2 [1 2 3])

# Arrays
(def array1 @(:one :two :three))
(def array2 @[:one :two :three])

# Structs
(def structs [
	{}
	{:key1 "value1" :key2 :value2 :key3 3}
	{(do 1 2 3) (do 4 5 6)}
	{@[] @[]}
	{1 2 3 4 5 6} ])

# Tables
(def structs [
	@{}
	@{:key1 "value1" :key2 :value2 :key3 3}
	@{(do 1 2 3) (do 4 5 6)}
	@{@[] @[]}
	@{1 2 3 4 5 6} ])

# Shorthand
(def x ())
(def shorthand [
	'x "Shorthand for (quote x)"
	;x "Shorthand for (splice x)"
	~x "Shorthand for (quasiquote x)"
	#,x "Shorthand for (unquote x)"
	|(print x) "Shorthand for (short-fn (body $))"
	"These shorthand notations can be combined in any order, allowing forms like"
	''x "((quote (quote x))), or"
	#,;x "((unquote (splice x)))"
	'def
	';quote
	])

# The janet source code at the following link is also a good test for syntax
# highlighting:
# https://github.com/janet-lang/janet/blob/master/src/boot/boot.janet
```

engenforge:
And here's a screen shot of syntax highlighting using the test file above and the [vis](https://github.com/martanne/vis) editor.  The colors are a bit gaudy, but it was helpful for me during syntax debugging to clearly differentiate the different pieces identified by the lexer.
20250621_19h15m01s_grim

orbitalquark:
Well done, thank you! I made some slight formatting and comment changes, and also cleaned up the string rule. https://github.com/orbitalquark/scintillua/commit/3ebcdabbe6758eb98853abfafc64320c1161b4dd. Let me know if I missed anything.

engenforge:
@orbitalquark, Thank you for merging this.  Your string rule is a great improvement.  I noticed that you had to remove the Lpeg utfR usage due to it being in the latest version.  As a result unicode identifiers aren't syntax highlighted in the Janet lexer.  Any suggestions for a workaround, or will that have to wait until the latest version of Lpeg is available?

orbitalquark:
How important is it to highlight unicode as an identifier as opposed to default? Personally, I'd expect legitimate use cases to be extremely low.

If it is important, perhaps we can get away with this:
```
local id_ch = S('!@$%^&*:-_+=<>.?') + lexer.alnum +
	(lpeg.utfR and lpeg.utfR(0x7F, 0x10FFFF) or lpeg.R('\128\255')^1)
```

What do you think?

engenforge:
I agree that the use cases are low.  It won't affect my personal use.  I grep'd the Janet repository for unicode identifiers and found a couple uses in the test suite, but otherwise it's pretty much non-existent.  I was mainly interested in it from the perspective of having the syntax lexer as accurate as possible.  I'm fine waiting rather than using a non-ideal solution in the short term.

Is the plan at some point in the future that scintillua would be able to handle syntax highlighting for unicode?  Would it be appropriate to open a low priority issue for unicode support that gets closed when Lpeg utfR is available?

orbitalquark:
Syntactically-significant unicode is very unusual in programming languages, so it's not a high priority to support it with a high level of accuracy (i.e. `lpeg.R('\128\255')` or similar patterns may be enough in certain contexts). Lexers are typically not designed to be parsers, so "just good enough" applies.

I don't have a future plan. A loose guideline for minimum dependencies is whatever the oldest supported Ubuntu LTS has in its package repository. It doesn't look like even Ubuntu 24.04 has lpeg 1.1.0 (which has `lpeg.utfR()`), so I don't expect to bump the lpeg 1.0.0 dependency for a long time.

engenforge:
Thank you for the reply.  That makes sense.

PR #163, opened 2025-06-17
fix(org): use better names for different stages of TODO items progress

### LINKED ISSUE #161: Better classification of the item status in the org lexer
Currently we have:

```lua
-- ToDos.
lex:add_rule('todo', lex:tag('TODO', lex:word_match('TODO')))
lex:add_rule('done', lex:tag('DONE', lex:word_match('DONE')))
lex:add_rule('wontfix', lex:tag('WONTFIX', lex:word_match('WONTFIX')))
```

(and the actual keywords which are matched by these rules can be changed later by a script)

After using this for a month or so (yikes, since February 5, time flies!), I think I got the classification wrong. I think the three levels, should be:

* `settled` for the states which are not complete but don’t require my immediate intervention (`TODO`, `DELEGATED`, etc.)
* `waiting` I am not done with my work for a day until the item moves out of this state
* `done` (which can include also `WONTFIX`, `INVALID`, etc.)

What do you think? Should I modify `org.lua` (and my theme files) accordingly?

mcepl:
I may fiddle a bit later with which exact keywords goes to which "bucket", but that’s done outside of the lexer itself, so I think, this is good.

Fixes: https://github.com/orbitalquark/scintillua/issues/161

### DIFF
--- a/lexers/org.lua
+++ b/lexers/org.lua
@@ -31,9 +31,9 @@ lex:add_rule('italic', lex:tag('ITALIC', '/' * lexer.word * '/'))
 lex:add_rule('underline', lex:tag('UNDERLINE', '_' * lexer.alnum * '_'))
 
 -- ToDos.
-lex:add_rule('todo', lex:tag('TODO', lex:word_match('TODO')))
+lex:add_rule('settled', lex:tag('SETTLED', lex:word_match('SETTLED')))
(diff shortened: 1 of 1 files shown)


orbitalquark:
Committed in https://github.com/orbitalquark/scintillua/commit/de2b8e4c289548f241e05d95aa7cd749a3f90deb. Thank you!

PR #169, opened 2025-07-10
Update dark.properties to have consistent color list as light

clach04:
Removed unused color, and resorted so light and dark color order is the same for easier comparisons/diffs.


orbitalquark:
Committed in https://github.com/orbitalquark/scintillua/commit/0908e54e48f2e5eba824851142907a30eea3d828. Thanks.

PR #170, opened 2025-07-11
Similar light dark theme for easier color updates

clach04:


clach04:
diff'ing light and dark now yields:

```diff
C:\code\scite\scintillua\themes>"C:\Program Files\Git\usr\bin\diff.exe" light.properties dark.properties
2c2
< # Light theme for Scintillua.
---
> # Dark theme for Scintillua.
5c5
< scintillua.colors.black=#1A1A1A
---
> scintillua.colors.black=#000000
25,26c25,26
< scintillua.colors.bg_0=$(scintillua.colors.white)
< scintillua.colors.bg_1=$(scintillua.colors.light_grey)
---
> scintillua.colors.bg_0=$(scintillua.colors.black)
> scintillua.colors.bg_1=$(scintillua.colors.light_black)
29c29
< scintillua.colors.fg_0=$(scintillua.colors.black)
---
> scintillua.colors.fg_0=$(scintillua.colors.light_grey)
```

I.e. only have to update colors, and maybe the foreground and background mapping. Can leave all the other settings alone without being forced to customize.

orbitalquark:
Sorry, I don't care for this change. If I need to change a color or experiment with colors, it's much easier for me to read and write actual color names instead of 'fg_0' and 'fg_1'. I'm not even going to remember which is which. And if I want to use something other than those colors, I may have to define 'fg_2', which will result in the same kind of problem.

clach04:
Gotcha. We're coming at this from different angles.

From the perspective of, a single theme, I think you would prefer to tweak that theme and colors.

I'm coming at this from the angle where I don't want to do that, I want to use a theme engine where there is an abstraction in place. Ala Base16, Base24, Catppuccin, Tomorrow, Solarized, etc.

I'll close this down and work on some theme frameworks (I mostly have Base16 working already, I need to open at least one ticket to discuss possible alternatives related to conditionals).

PR #176, opened 2025-08-19
Added crontab lexer

klokocka:
Basic crontab file format lexer (https://en.wikipedia.org/wiki/Cron).
Tested and working on my local computer.

### DIFF
--- a/lexers/crontab.lua
+++ b/lexers/crontab.lua
@@ -4,7 +4,7 @@
 local lexer = require('lexer')
 local P, S, R = lpeg.P, lpeg.S, lpeg.R
 
-local lex = lexer.new(...)
+local lex = lexer.new(..., {no_user_word_lists = true})
 
 -- Define basic patterns
 local space = S(' \t')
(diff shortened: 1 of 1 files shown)


orbitalquark:
Thanks for starting on this! This lexer is written in the legacy format, so if you want this to be merge, you'll need to migrate it using the steps outlined here: https://orbitalquark.github.io/scintillua/api.html#migrating-legacy-lexers

The styles will probably be tricky. For the minutes, hours, etc., I would suggest you use tags like `lex:tag(lexer.NUMBER .. '.minute')`, `lex:tag(lexer.NUMBER .. '.hour')`, etc. Then theme files can add separate foreground colors for those categories should they choose to.

I can do a more thorough review once the migration is complete.

If you'd prefer not to put in the work, that's fine. We can just leave this PR up for anyone interested in using it as-is.

klokocka:
Updated and simplified...

orbitalquark:
No, that migration does not look correct. If you used an LLM, I discovered they aren't very good following the migration steps :(

PR #181, opened 2025-10-01
lexer.lua: detect() Merging vis filetype detection

Nomarian:
First stage in merging vis file detection to textadept

Many editors create backup files, vim creates ~, others create .bak and so on, this ignores and gets the true filetype

### DIFF
--- a/lexers/lexer.lua
+++ b/lexers/lexer.lua
@@ -1676,20 +1676,106 @@ M.ignored_extensions = {
 	orig = true, bak = true, old = true, new = true
 }
 
+--- Map of hashbang utilities to associated lexer names.
+-- @usage lexer.detect_utilities.luadoc = 'lua'
+M.detect_utilities = {}
+
 --- Map of first-line patterns to their associated lexer names.
 -- These are Lua string patterns, not LPeg patterns.
 -- @usage lexer.detect_patterns['^#!.+/zsh'] = 'bash'
 M.detect_patterns = {}
 
---- Returns the name of the lexer often associated a particular filename and/or file content.
--- @param[opt] filename String filename to inspect. The default value is read from the
---   "lexer.scintillua.filename" property.
--- @param[optchain] line String first content line, such as a shebang line. The default value
---   is read from the "lexer.scintillua.line" property.
--- @return string lexer name to pass to `lexer.load()`, or `nil` if none was detected
-function M.detect(filename, line)
-	if not filename then filename = M.property and M.property['lexer.scintillua.filename'] or '' end
-	if not line then line = M.property and M.property['lexer.scintillua.line'] or '' end
+--[[ hashbang check
+	hashbangs only have PATH <SPACE> ARG
+	NOTE: #! have a line limit of 256 characters
+		if PATH matches /env, utility should be in args
+			POSIX does not have -S, but BSDs/Linux seem to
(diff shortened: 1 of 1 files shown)


lexers/lexer.lua
```diff
@@ -1676,20 +1676,106 @@ M.ignored_extensions = {
 	orig = true, bak = true, old = true, new = true
 }
 
+--- Map of hashbang utilities to associated lexer names.
+-- @usage lexer.detect_utilities.luadoc = 'lua'
+M.detect_utilities = {}
```
orbitalquark, line 1681: What is the difference between this and `M.detect_patterns`?

lexers/lexer.lua
```diff
@@ -1676,20 +1676,106 @@ M.ignored_extensions = {
...
+	hashbangs only have PATH <SPACE> ARG
+	NOTE: #! have a line limit of 256 characters
+		if PATH matches /env, utility should be in args
+			POSIX does not have -S, but BSDs/Linux seem to
+			All do seem to have name=value
+			discard first arg if /-[^S]*S/; and all subsequent /=/
+			NOTE: this means you can't have a command with /^-|=/
+	NOTES: long-options (GNUisms) unsupported
+--]]
+local function GetHashBang(line)
```
orbitalquark, line 1698: This seems like an awful lot of work to avoid populating `M.detect_patterns` with more hashbang patterns.

lexers/lexer.lua
```diff
@@ -1848,41 +1934,31 @@ function M.detect(filename, line)
...
-			or extensions[name]
-			or M.detect_extensions[ext]
-			or extensions[ext]
+	local lower = ext and ext:lower()
+	return M.detect_extensions[name]
+		or extensions[name]
+		or M.detect_extensions[ext]
+		or extensions[ext]
+		or M.detect_extensions[lower]
+		or extensions[lower]
```
orbitalquark, line 1944: I'm not entirely sure how I feel about this. Filenames are case-sensitive on Linux for example, so some applications might not want to auto-lowercase extensions for fear of false-detections. On the other hand, applications that are on case-insensitive filesystems may simply want to pass a lowercase filename to `M.detect()` to get the desired behavior.

orbitalquark (commented):
I like having the ability to ignore certain file suffixes, but I'm not sure about the rest of it. I've left some comments to think about.

lexers/lexer.lua
Nomarian: Given that env -S allows for variables, you could have a variable called { env -S python_ignore=true awk -f } would lead to python being set as the syntax, a false positive.

lexers/lexer.lua
Nomarian: I don't like it either but I was tired of opening old .XML files and it not being identified properly

lexers/lexer.lua
Nomarian: POSIX calls utility the part of env which is the launcher, so in { env -S awk -f } awk would be the utility.

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/env.html

Nomarian:
There's still some things yet to be ported, vis opens file --mime-type as a last resort if everything fails which I don't know if I should add

Nomarian:
This merge is too big, I'll split it into multiple pull requests instead.

PR #182, opened 2025-10-14
Add class/constructor detection rule for Dart

M0JXD:
Hi,

I've been playing with this some more and comparing to other editors highlighting, and I've (or well, mostly ChatGPT...) added a rule to highlight classes/constructors. The [Dart Style Guide](https://dart.dev/effective-dart/style) is highly enforced by Dart's LSP so UpperCamelCase words are guaranteed to be a class or constructor.

Thanks,

Jamie

Edit: I just want to clarify that I mostly try not to use ChatGPT/AI, and my previous contributions have been all me with no AI input. But when it comes to Regex and Pattern Matching I'm hapless 😅. Thankfully I did learn a bit about LPeg after I had it explain why it works!

### DIFF
--- a/lexers/dart.lua
+++ b/lexers/dart.lua
@@ -23,6 +23,12 @@ local dq_str = S('r')^-1 * lexer.range('"', true)
 local tq_str = S('r')^-1 * (lexer.range("'''") + lexer.range('"""'))
 lex:add_rule('string', lex:tag(lexer.STRING, tq_str + sq_str + dq_str))
 
+-- Capitalized identifiers (likely classes or constructors).
+local upper = lpeg.R('AZ')
+local alnum = lpeg.R('AZ', 'az', '09') + P('_')
+local capitalized_word = (lexer.word_boundary or P(true)) * upper * alnum^0
+lex:add_rule('constructor', lex:tag(lexer.TYPE, capitalized_word))
+
 -- Functions.
 lex:add_rule('function', lex:tag(lexer.FUNCTION, lexer.word) * #(lexer.space^0 * '('))
 
--- a/lexers/dart.lua
+++ b/lexers/dart.lua
@@ -12,8 +12,13 @@ local lex = lexer.new(...)
 lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD)))
 -- Built-ins.
 lex:add_rule('constant', lex:tag(lexer.CONSTANT_BUILTIN, lex:word_match(lexer.CONSTANT_BUILTIN)))
--- Types.
-lex:add_rule('type', lex:tag(lexer.TYPE, lex:word_match(lexer.TYPE)))
+
+-- Types (also matches capitalised words)
+local upper = lpeg.R('AZ')
+local alnum = lpeg.R('AZ', 'az', '09') + P('_')
+local capitalized_word = upper * alnum^0
+lex:add_rule('type', lex:tag(lexer.TYPE, lex:word_match(lexer.TYPE) + capitalized_word))
+
 -- Directives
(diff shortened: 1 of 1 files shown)


lexers/dart.lua
```diff
@@ -23,6 +23,12 @@ local dq_str = S('r')^-1 * lexer.range('"', true)
 local tq_str = S('r')^-1 * (lexer.range("'''") + lexer.range('"""'))
 lex:add_rule('string', lex:tag(lexer.STRING, tq_str + sq_str + dq_str))
 
+-- Capitalized identifiers (likely classes or constructors).
+local upper = lpeg.R('AZ')
+local alnum = lpeg.R('AZ', 'az', '09') + P('_')
+local capitalized_word = (lexer.word_boundary or P(true)) * upper * alnum^0
```
orbitalquark: You can delete the `(lexer.word_boundary or P(true))` bit because `lexer.word_boundary` does not exist.

This pattern will match something like `R_1`, `Snake_case`, and `SCREAMING_CAPS`. If you only want to match CamelCase, then you should omit the `+ P('_')` part from `alnum`.

lexers/dart.lua
```diff
@@ -23,6 +23,12 @@ local dq_str = S('r')^-1 * lexer.range('"', true)
...
+lex:add_rule('constructor', lex:tag(lexer.TYPE, capitalized_word))
```
orbitalquark: This feels more like an extension of the existing 'type' rule above. It's not a constructor, as it will tag the 'Foo' in `class Foo`, `typedef Foo = Bar`, etc. Maybe we should put it up there.

orbitalquark (changes requested):
I'm not sure how I feel about this. On the one hand it will match more generic types, but on the other it may be too overzealous for code that doesn't follow the style guide to the letter.

Something particularly troublesome is that the style guide says "PREFER using lowerCamelCase for constant names". It's not a "DO". This PR will tag `const ID = "foo"` as a type, which is not correct.

lexers/dart.lua
M0JXD: > I'm not sure how I feel about this. On the one hand it will match more generic types, but on the other it may be too overzealous for code that doesn't follow the style guide to the letter.

Yes I understand the reluctance. It's not consistent across other editors likely for the same reasons. As you can see here by default VSCode (the most used Dart Editor which I'm aiming to match), Helix and Xed have this kind of highlighting but Geany and Kate do not.

Other Editors

<br><br>

> You can delete the (lexer.word_boundary or P(true)) bit because lexer.word_boundary does not exist.

Well that teaches me for believing anything AI says! If you want this merged then I'll have look at this properly and clean it up, I think you're right it's more of an extension of the type rule than a class/constructor rule.

> This pattern will match something like R_1, Snake_case, and SCREAMING_CAPS

This was intended, as this is how VSCode behaves:

Vscode highlights these

M0JXD:
Hi, 

I've improved this a bit more (why was I so scared of learning LPeg? It's awesome!) and also added detection of capitalised words with a leading underscore, which is common with class names for Flutter's [Stateful Widgets](https://docs.flutter.dev/get-started/fundamentals/widgets#widget-state).

Thank you for your help and patience 🙂

orbitalquark:
Thanks, I've simplified your PR into this commit: https://github.com/orbitalquark/scintillua/commit/68eed24699f42478a986faa1dd2470e72d273f1c.

I'm not crazy about highlighting a potential constant (e.g. `const ID = 1`) as a type, but oh well. Hopefully it's not too common.

M0JXD:
Thank you. I've checked and the lexer still seems to behave if I set it as a line lexer, so maybe something could be done to detect if `const` is in the current line and then not apply the caps highlighting. I'll look into it at some point. As of now it behaves the same as other editors.

Thank you again.

PR #184, opened 2025-12-26
Add keyword and functions from Lua 5.5

nthirtyone:
Diff of relevant part of reference manuals:

```diff
--- 5.4
+++ 5.5
@@ -92,7 +92,9 @@
 math.exp
 math.floor
 math.fmod
+math.frexp
 math.huge
+math.ldexp
 math.log
 math.max
 math.maxinteger
@@ -154,6 +156,7 @@

 table
 table.concat
+table.create
 table.insert
 table.move
 table.pack
```

and syntax

```diff
--- 5.4
+++ 5.5
@@ -16,11 +16,14 @@ stat ::=
     for namelist in explist do block end |
     function funcname funcbody |
     local function Name funcbody |
-    local attnamelist [‘=’ explist]
+    global function Name funcbody |
+    local attnamelist [‘=’ explist] |
+    global attnamelist |
+    global [attrib] ‘*’

-attnamelist ::=  Name attrib {‘,’ Name attrib}
+attnamelist ::=  [attrib] Name [attrib] {‘,’ Name [attrib]}

-attrib ::= [‘<’ Name ‘>’]
+attrib ::= ‘<’ Name ‘>’

 retstat ::= return [explist] [‘;’]

@@ -49,7 +52,9 @@

 funcbody ::= ‘(’ [parlist] ‘)’ block end

-parlist ::= namelist [‘,’ ‘...’] | ‘...’
+parlist ::= namelist [‘,’ varargparam] | varargparam
+
+varargparam ::= ‘...’ [Name]

 tableconstructor ::= ‘{’ [fieldlist] ‘}’

```

I feel like we can leave `...abc` as it is: `abc` acts as an identifier and is matched as such; `...` is left as it was. As for `'global' [attrib] '*'`, `*` is matched as an operator which feels wrong but I'm unsure which one would be more appropriate (identifier? hardly; keyword?).

GitHub decided to switch display names for no reason (again?), seems I'm already in [thanks.md](https://github.com/orbitalquark/scintillua/blob/default/docs/thanks.md?plain=1#L8) so avoid duplicating with different name in case you choose to use this patch.

### DIFF
--- a/lexers/lua.lua
+++ b/lexers/lua.lua
@@ -81,7 +81,8 @@ lex:add_fold_point(lexer.OPERATOR, '{', '}')
 lex:set_word_list(lexer.KEYWORD, {
 	'and', 'break', 'do', 'else', 'elseif', 'end', 'false', 'for', 'function', 'if', 'in', 'local',
 	'or', 'nil', 'not', 'repeat', 'return', 'then', 'true', 'until', 'while', --
-	'goto' -- 5.2
+	'goto', -- 5.2
+	'global' -- 5.5
 })
 
 lex:set_word_list(lexer.FUNCTION_BUILTIN, {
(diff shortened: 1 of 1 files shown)


orbitalquark:
Thanks for bringing this up. I had been working on this prior to 5.5's release. I've made the changes in https://github.com/orbitalquark/scintillua/commit/3847bbf66dc2762344427eebe5fe172eb2d349f4.

I agree that '*' as an operator in a global statement is not ideal, but oh well.

`math.frexp()` and `math.ldexp()` existed in Lua 5.3, so I don't feel the need to mark them as 5.5.

PR #185, opened 2026-01-11
Add semantic token subtypes to Python lexer

nishantvarma:
This PR adds semantic token subtypes to the Python lexer for more granular syntax highlighting:

  **New subtypes:**
  - `KEYWORD.declaration` — `def`, `class`, `lambda`
  - `KEYWORD.import` — `import`, `from`
  - `CONSTANT_BUILTIN.boolean` — `True`, `False`, `None`
  - `STRING.doc` — triple-quoted strings (docstrings)
  - `VARIABLE_BUILTIN` — `self`, `cls`

  Enables editors/themes to visually distinguish:
  - Creation keywords from control flow
  - Imports from other keywords
  - Boolean literals from generic constants
  - Docstrings from regular strings
  - Instance references from regular identifiers

  **Also fixes:** duplicate `'Exception'` in CONSTANT_BUILTIN word list.

### DIFF
--- a/lexers/python.lua
+++ b/lexers/python.lua
@@ -8,12 +8,23 @@ local P, S, B = lpeg.P, lpeg.S, lpeg.B
 local lex = lexer.new(..., {fold_by_indentation = true})
 
 -- Classes.
-lex:add_rule('classdef', lex:tag(lexer.KEYWORD, 'class') * lex:get_rule('whitespace') *
-	lex:tag(lexer.CLASS, lexer.word))
+lex:add_rule('classdef', lex:tag(lexer.KEYWORD .. '.declaration', 'class') *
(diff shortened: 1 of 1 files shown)


orbitalquark:
Thanks for submitting this. I've fixed the duplicate `Exception` string in https://github.com/orbitalquark/scintillua/commit/a739b821817c6810ed50665c6022e0261d49a3cb.

As for these extra token types, I feel this is too subjective to be included by default for just this lexer. If I included them, I'd like to do something similar for other lexers so that assigning a color to `lexer.KEYWORD..'.import'`, for example, would affect Java imports, Ruby requires, etc. This would require a more general set of sub-tags, and a decent amount of effort to determine what they should be.

PR #186, opened 2026-02-10
Ignores patterns when detecting lexer

Nomarian:
Many editors create backup files, ~ .bak .new, this takes care of that (Taken from vis)


orbitalquark:
Thanks for putting this together. I've committed a different implementation of this feature in https://github.com/orbitalquark/scintillua/commit/6abc41aee41791c9e124f6e7bc09f0ba5c2a11d0.

PR #187, opened 2026-02-16
Merging vis filetype.lua detections (that match lexer)

Nomarian:
This probably won't be merged, and I don't know if any of these extensions really are what they say they are, its just there, so I think it should be here instead.

Missing ones were for modula, asciidoc, typst, git-commit (diff?), mail, usfm

I don't like to mix up filenames and extensions, which is why its separate, extensions usually are case insensitive. aanyway.

### DIFF
--- a/lexers/lexer.lua
+++ b/lexers/lexer.lua
@@ -1715,6 +1715,7 @@ function M.detect(filename, line)
 		elm = 'elm', --
 		erl = 'erlang', hrl = 'erlang', --
 		fs = 'fsharp', --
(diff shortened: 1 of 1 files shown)


orbitalquark:
I've committed some of these in https://github.com/orbitalquark/scintillua/commit/c48982c274b9c01f9c0a91486b10a0bd385274e2. I'm more comfortable with using "official" extensions rather than conventional ones.

Nomarian:
The filenames are pretty official, like [".login"] = "bash"

PR #188, opened 2026-02-16
Add hashbang parser function to get utility directly and use a lookup table

Nomarian:


orbitalquark:
Sorry, I don't see the utility in something like this. `lexer.detect_patterns` is pretty explicit and I think we should continue to use it rather than having another layer on top. Scintillua clients can use a layer like this to generate patterns for `lexer.detect_patterns`.

PR #191, opened 2026-03-10
Recognise Arduino and 'H' extensions

M0JXD:
Very small PR to add these :)


orbitalquark:
Thanks! Committed in https://github.com/orbitalquark/scintillua/commit/c58ef40d1544ff573414d08bc8a09741a70f30dd

orbitalquark:
The Arduino Sketch language is separate from C and C++ and should probably be its own lexer. Otherwise, clients that want to highlight it as C or C++ can use `lexer.detect_extensions.ino = 'c'` and perhaps even edit the word lists after calling `lexer.load()`.

PR #192, opened 2026-03-14
Arduino Lexer

M0JXD:
As said in #191, just starting a new lexer for this and opening here as WIP.

As I see it Arduino is a C++ library with a smidge of preprocessing, so I've inherited the C++ lexer and added what's in the [language reference](https://docs.arduino.cc/language-reference/).

I'm unsure whether `setup()` and `loop()` should be lexed as you don't lex `main()` in other languages, and they're basically Arduino's version of the main entry point.

Next thing I'll do is get the common base members to be recognised as builtin functions for SPI, Wire, Serial etc., which I think is best done the same way you handle the `std::` prefix in the C++ lexer so I'll modify the `'function'` rule.

### DIFF
--- a/lexers/arduino.lua
+++ b/lexers/arduino.lua
@@ -74,7 +74,7 @@ lex:set_word_list(lexer.CONSTANT_BUILTIN,
 lex:set_word_list(lexer.VARIABLE_BUILTIN, {
 	'EEPROM', 'SPI', 'Wire', 'Serial', 'Mouse', 'Keyboard',	'WiFi', 'BLE',
 	'LiquidCrystal', 'lcd'
-}, true)
+})
 
 lexer.property['scintillua.comment'] = '//'
 
--- a/lexers/arduino.lua
+++ b/lexers/arduino.lua
@@ -1,4 +1,4 @@
--- Arduino LPeg lexer.
+-- Arduino LPeg lexer. Mostly copied from C++ lexer (preamble below)
 -- Reference: https://docs.arduino.cc/language-reference/
 
 -- Copyright 2006-2026 Mitchell. See LICENSE.
--- a/lexers/arduino.lua
+++ b/lexers/arduino.lua
@@ -5,75 +5,76 @@ local lexer = lexer
 local P, S, B = lpeg.P, lpeg.S, lpeg.B
 local lex = lexer.new(..., {inherit = lexer.load('c')})
 
--- C++ member rule that also allows builtins to be highlighted
-local non_member = -(B('.') + B('->') + B('::'))
-local builtin_func = lex:tag(lexer.FUNCTION_BUILTIN, lex:word_match(lexer.FUNCTION_BUILTIN))
-local func = lex:tag(lexer.FUNCTION, lexer.word)
-local method = (B('.') + B('->')) * lex:tag(lexer.FUNCTION_METHOD, lexer.word)
-lex:modify_rule('function', ((builtin_func) * non_member + method + func) * #(lexer.space^0 * '('))
-
-lex:modify_rule('constants', lex:get_rule('constants') +
-	lex:tag(lexer.VARIABLE_BUILTIN, lex:word_match(lexer.VARIABLE_BUILTIN)))
-
 lex:set_word_list(lexer.KEYWORD, {
 	-- Additional C++ Keywords (they highlight in Arduino IDE v1.8)
-	'catch', 'class', 'const_cast',	'delete', 'dynamic_cast', 'explicit', 'export', 'friend',
-	'mutable', 'namespace', 'new', 'operator', 'private', 'protected', 'public',
-	'reinterpret_cast', 'static_cast', 'template', 'this', 'throw', 'try',
-	'typeid', 'typename', 'using', 'virtual',
-	'and', 'not', 'or', 'xor', 	-- Operators
+	'catch', 'class', 'const_cast', 'delete', 'dynamic_cast', 'explicit', 'export', 'friend',
+	'mutable', 'namespace', 'new', 'operator', 'private', 'protected', 'public', 'reinterpret_cast',
+	'static_cast', 'template', 'this', 'throw', 'try', 'typeid', 'typename', 'using', 'virtual',
+	'and', 'not', 'or', 'xor', -- Operators
 	'final', 'override', -- C++11.
 	'PROGMEM' -- Arduino
 }, true)
 
-lex:set_word_list(lexer.TYPE, {
-	-- Additional C++ Types (they highlight in Arduino IDE v1.8)
-	'wchar_t', --
-	'char16_t', 'char32_t', -- C++11
-	-- <cstddef>
-	'byte', -- C++17
-	-- Arduino
-	'word', 'String'
-}, true)
+lex:set_word_list(lexer.TYPE, 'byte word String', true)
 
 lex:set_word_list(lexer.FUNCTION_BUILTIN, {
 	-- I/O
-	'digitalRead', 'digitalWrite', 'pinMode',
-	'analogRead', 'analogReadResolution', 'analogReference',
-	'analogWrite', 'analogWriteResolution',
+	'digitalRead', 'digitalWrite', 'pinMode', 'analogRead', 'analogReadResolution',
+	'analogReference', 'analogWrite', 'analogWriteResolution',
 	-- Adv I/O
 	'noTone', 'pulseIn', 'pulseInLong', 'shiftIn', 'shiftOut', 'tone',
 	-- Time
 	'delay', 'delayMicroseconds', 'micros', 'millis',
-	-- Maths
+	-- Maths (most are covered by cmath, just adding the missing ones)
 	'abs', 'constrain', 'map', 'max', 'min', 'pow', 'sq', 'exp', 'bit',
 	-- Characters
-	'isAlpha', 'isAlphaNumeric', 'isAscii', 'isControl', 'isDigit', 'isGraph',
-	'isHexadecimalDigit', 'isLowerCase', 'isPrintable', 'isPunct', 'isSpace',
-	'isUpperCase', 'isWhitespace',
+	'isAlpha', 'isAlphaNumeric', 'isAscii', 'isControl', 'isDigit', 'isGraph', 'isHexadecimalDigit',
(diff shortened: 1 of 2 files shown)


M0JXD:
The function rule is changed to allow the builtin names to highlight when they are member functions.

I've added some common peripheral classes under VARIABLE_BUILTIN (which seemed most appropriate), but they can't be lexed as the inherited `'identifier'` rule from the C++ lexer is taking priority. I tried using `modify_rule` but that doesn't change rule order.

Is there solution for when adding a new class of "keywords" to an inherited lexer to get around nearly every lexer's `lexer.word` identifier rule? At the moment it seems the only option is to copy and modify the C++ lexer instead of inheriting.

orbitalquark:
> I've added some common peripheral classes under VARIABLE_BUILTIN (which seemed most appropriate), but they can't be lexed as the inherited `'identifier'` rule from the C++ lexer is taking priority. I tried using `modify_rule` but that doesn't change rule order.
> 
> Is there solution for when adding a new class of "keywords" to an inherited lexer to get around nearly every lexer's `lexer.word` identifier rule? At the moment it seems the only option is to copy and modify the C++ lexer instead of inheriting.

Maybe you can `lex:modify_rule('constants', lex:get_rule('constants') + lex:word_match(lexer.VARIABLE_BUILTIN))`?

I think that's similar to what other lexers are doing.

M0JXD:
> Maybe you can lex:modify_rule('constants', lex:get_rule('constants') + lex:word_match(lexer.VARIABLE_BUILTIN))?

No such luck 🙁

I've changed it to copied from the C++ lexer if that is okay by you?

orbitalquark:
No, based on my understanding of the language spec, you should not be copying the C++ lexer with its STL into an Arduino lexer. Ideally, I think it would probably be better to inherit from the C lexer since C++ has extra language features like attributes, string prefixes, etc. that Arduino doesn't have. I may investigate the inheritance rule issue you experienced when I have some time.

M0JXD:
As an idea, the inheritance issue could be fixed by adding a variable_builtin rule to the C++ lexer prior to the identifier rule without a word list, so any lexer that inherits it has the option to use the tag by extending the word list:

`lex:add_rule('variable_builtin', lex:tag(lexer.VARIABLE_BUILTIN, lex:word_match(lexer.VARIABLE_BUILTIN)))`

> Ideally, I think it would probably be better to inherit from the C lexer since C++ has extra language features like attributes, string prefixes, etc. that Arduino doesn't have.

Those aren't really purposeful language omissions applied by Arduino, but of [avr-g++](https://www.microchip.com/en-us/tools-resources/develop/microchip-studio/gcc-compilers) which can only implement a subset of C++11 for 8bit uCs. E.g. if someone was using [PlatformIO](https://github.com/platformio/platform-atmelavr/tree/master/examples/arduino-blink) or [Microchip Studio](https://www.microchip.com/en-us/tools-resources/develop/microchip-studio) they would make *.cpp* files, and end up with the C++ lexer in a Scintillua editor despite the same limitations of avr-g++ in those situations. Even if you're sticking within the Arduino ecosystem if you start writing libraries [you must use the .cpp extension](https://docs.arduino.cc/learn/contributions/arduino-creating-library-guide/).

M0JXD:
Hi again, I've adjusted it to inherited from the C lexer as you suggested.
I also worked out the issue with the modified constants rule, we forgot to specify the new tag. This works: 

`lex:modify_rule('constants', lex:get_rule('constants') + lex:tag(lexer.VARIABLE_BUILTIN, lex:word_match(lexer.VARIABLE_BUILTIN)))`

Let me know any thoughts.

orbitalquark:
Thanks for sticking with it. This looks a lot better. I'll have another look in the coming days.

lexers/arduino.lua
```diff
@@ -0,0 +1,81 @@
+-- Arduino LPeg lexer.
+-- Reference: https://docs.arduino.cc/language-reference/
+
+local lexer = lexer
+local P, S, B = lpeg.P, lpeg.S, lpeg.B
+local lex = lexer.new(..., {inherit = lexer.load('c')})
+
+-- Modify to allow builtins to be highlighted even as class members
+local non_member = -(B('.') + B('->') + B('::'))
```
orbitalquark: We don't need this because Arduino does not have structs.

lexers/arduino.lua
```diff
@@ -0,0 +1,81 @@
...
+local builtin_func = lex:tag(lexer.FUNCTION_BUILTIN, lex:word_match(lexer.FUNCTION_BUILTIN))
+local func = lex:tag(lexer.FUNCTION, lexer.word)
+local method = (B('.') + B('->')) * lex:tag(lexer.FUNCTION_METHOD, lexer.word)
```
orbitalquark: Ditto, we don't need this because Arduino doesn't have structs.

lexers/arduino.lua
```diff
@@ -0,0 +1,81 @@
...
+lex:modify_rule('function', ((builtin_func) * non_member + method + func) * #(lexer.space^0 * '('))
+
+lex:modify_rule('constants', lex:get_rule('constants') +
+	lex:tag(lexer.VARIABLE_BUILTIN, lex:word_match(lexer.VARIABLE_BUILTIN)))
+
+lex:set_word_list(lexer.KEYWORD, {
+	-- Additional C++ Keywords (they highlight in Arduino IDE v1.8)
```
orbitalquark, line 9: I'm not seeing any of these in the language reference. Is it possible Arduino IDE is using a stock C++ lexer instead of one that fits their dialect? Can an Arduino script compile with any of these C++ keywords in it?

I think I would prefer seeing just the set of keywords specified in the language reference.

lexers/arduino.lua
```diff
@@ -0,0 +1,81 @@
...
+	'catch', 'class', 'const_cast',	'delete', 'dynamic_cast', 'explicit', 'export', 'friend',
+	'mutable', 'namespace', 'new', 'operator', 'private', 'protected', 'public',
+	'reinterpret_cast', 'static_cast', 'template', 'this', 'throw', 'try',
+	'typeid', 'typename', 'using', 'virtual',
+	'and', 'not', 'or', 'xor', 	-- Operators
+	'final', 'override', -- C++11.
+	'PROGMEM' -- Arduino
+}, true)
+
+lex:set_word_list(lexer.TYPE, {
```
orbitalquark: Again, I'd probably just stick to the types mentioned in the language reference.

lexers/arduino.lua
```diff
@@ -0,0 +1,81 @@
...
+	-- Additional C++ Types (they highlight in Arduino IDE v1.8)
+	'wchar_t', --
+	'char16_t', 'char32_t', -- C++11
+	-- <cstddef>
+	'byte', -- C++17
+	-- Arduino
+	'word', 'String'
+}, true)
+
+lex:set_word_list(lexer.FUNCTION_BUILTIN, {
```
orbitalquark, line 20: Can we make it like the Lua lexer and include complete functions like `Serial.print`, `Serial.println`, `Stream.available`, `Stream.read`, etc.?

As written, a bare function named "read" would be recognized as a built-in function even though it's not -- it belongs to `Stream` for example.

orbitalquark (changes requested):

lexers/arduino.lua
M0JXD: Yes that would be better, I'll do that. It would also mean some of the members not in the abstract class can be included too.

lexers/arduino.lua
M0JXD: Arduino does have structs/classes but given we will change to using the whole names for function_builtin it can be removed.

lexers/arduino.lua
M0JXD: As I'm trying to say, Arduino is very close to standard C++11.
E.g., this silly example with a template, namespace and class I made compiles and runs on an Uno:

```cpp
template <typename T> T myMax(T x, T y){
  return (x > y) ? x : y;
}

namespace SerialUtils {
  void start() {
    Serial.println("THIS IS IN A NAMESPACE");
    // Needs a symbol at the end to tell parseInt to stop
    Serial.println("Enter in format: Compare 3 and 4!");
  }
}

class TemplateRunner {
  public:
  void intIt(int in1, int in2) {
    Serial.print("Running with ints... ");
    Serial.println(myMax(in1, in2));
  }

  void floatIt(int in1, int in2) {
    Serial.print("Running with floats... ");
    float f1 = float(in1);
    float f2 = float(in2);
    Serial.println(myMax(f1, f2));
  }

  void wordIt(int in1, int in2) {
    Serial.print("Running with words... ");
    word w1 = word(in1);
    word w2 = word(in2);
    Serial.println(myMax(w1, w2));
  }
};

void setup() {
  Serial.begin(9600);
  Serial.setTimeout(10000);
  SerialUtils::start();
}

void loop() {
  if (Serial.available() > 3) {
    int inByte1 = Serial.parseInt();
    int inByte2 = Serial.parseInt();
    TemplateRunner myRunner;
    myRunner.intIt(inByte1, inByte2);
    myRunner.floatIt(inByte1, inByte2);
    myRunner.wordIt(inByte1, inByte2);
  }
}
```

All the Arduino preprocessing really does is put setup and loop and put into a main, and places setup, loop and main at the end so you don't have to worry about function prototypes etc..

```cpp
int main() {
  setup();
  while(1) loop();
}
```

lexers/arduino.lua
M0JXD: They do work but given everyone just uses the types in the Arduino reference I'll remove them.

M0JXD:
I've made some changes as requested for the builtin functions.

I removed LCD as I think the main aim would be to provide a lexer that highlights functionality available and builtin to the board - e.g. EEPROM, WiFi etc. should be included as those peripherals are on Arduino boards, but libraries for Motors and LCDs etc. require external parts (and are really using the interfaces like GPIO, I2C etc. underneath anyways) so should not be treated as "builtin".

I haven't removed the additional C++ keywords yet. I make use of namespaces and structs/classes in my sketches and see several others that do too.

orbitalquark:
Thank you very much. Committed in https://github.com/orbitalquark/scintillua/commit/3ccff05505474960ebacde7dbf91e2f9694b7d78.

PR #195, opened 2026-03-21
Improve Perl lexing

1-1sam:
This PR does the following:
* Recognize `PL` files as Perl files.
* Add missing flags to regex operators.
* Recognize regexes that are followed by commas, so that `split` regexes are properly lexed.
* Add missing keywords/builtins.
* Lex `&` (subroutine references) and `*` (type globs) sigils.


orbitalquark:
Thank you! Committed in https://github.com/orbitalquark/scintillua/commit/99125233ed210de1f94b0a71aceb5ca739c8ddbf.

PR #199, opened 2026-04-17
Add Asciidoc lexer

M0JXD:
Picking up from #151

Marking as WIP because it still needs work, but opening an avenue for feedback.
So far:

- Correct heading rule as Asciidoc uses `=` instead of `#`
- Add [Optional Title rule](https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#more-delimited-blocks)
- Change strong/em rules as they are not the same as Markdown, changed code rules to be a 'monospace' rule
- Remove HTML embedding as Asciidoc is its own thing, not a HTML superset like Markdown 
- Improve Admonitions as the keywords should only be highlighted in certain scenarios
- Correct [list rule](https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#lists)
- Correct Horizontal Rule rule (Asciidoc uses `'''`)

There is the question of whether Asciidoc's (very limited) [Markdown compatibility](https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#markdown-compatibility) should be implemented. Not all implementations of Asciidoc support this. At the moment I have kept that compatibility in the Horizontal Rule rule but I am tempted to remove it.

Also mentioning #104

### DIFF
--- a/lexers/asciidoc.lua
+++ b/lexers/asciidoc.lua
@@ -1,0 +1,96 @@
+-- Copyright 2025 Matěj Cepl. See LICENSE.
+-- Copyright 2026 Jamie Drinkell. See LICENSE
+-- Asciidoc LPeg lexer.
+
+local lexer = lexer
+local P, S, B = lpeg.P, lpeg.S, lpeg.B
+
+-- TODO: Asciidoctor notes *OPTIONAL* Markdown compatibility
+-- https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#markdown-compatibility
+-- Do we inherit? It only works with asciidoctor, so not all implementations.
+
+local lex = lexer.new(...)
+
+-- Admonitions.
+lex:add_rule('admonition', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD) * #S(':')))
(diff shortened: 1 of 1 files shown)


M0JXD:
I've been looking at other editors with Asciidoc highlighting and most of them seem to include the optional Markdown support, so I've added the header rule. Also made a fix for when the header rule shouldn't be lexing lines of "=".

I think maybe the flanked range rule can be replaced with a standard `lexer.range` as bold/italic don't use the same delimiting character.

M0JXD:
I'm not going to be using Asciidoc, but have made the executive decision not to support Markdown in this lexer as there are a few implementations about that don't have that compatibility. 

There's a few things not detected (e.g. tables), but tables aren't detected in the Markdown lexer either. For now this matches VSCode's highlighting pretty well which seems to be "the standard" for Asciidoc.

I'm marking this ready for review, if someone wants to chip in and add Markdown compatibility because they need it they're welcome too, but given I've had no feedback I'm hoping this is good enough. I hope this resolves the "should Asciidoc inherit Markdown" problem because this is now a very different lexer.

orbitalquark:
Thanks for working on this. I'll review when I have some time.

PR #202, opened 2026-05-01
Added (very basic and jank) C3 lexer

nubbsterr:
Based off of the C lexer with some tweaks. I took most of the keywords/grammar from the [C3 implementation details section](https://c3-lang.org/implementation-details/grammar/), with the rest coming from my own experience with the language. I haven't used C3 much so I might have missed some stuff but this is a start.

Big thing to note is that this was created during the 0.7.x release (0.7.11 to be exact), so anybody in the future checking this out should take that into account for inevitably broken/useless keywords.

### DIFF
--- a/lexers/c3.lua
+++ b/lexers/c3.lua
@@ -4,63 +4,23 @@
 local lexer = lexer
 local P, S, B = lpeg.P, lpeg.S, lpeg.B
 
-local lex = lexer.new(...)
-
--- Keywords.
-lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD)))
-
--- Types.
-lex:add_rule('type', lex:tag(lexer.TYPE, lex:word_match(lexer.TYPE)))
-
--- Functions.
-local builtin_func = -(B('.') + B('->')) *
-	lex:tag(lexer.FUNCTION_BUILTIN, lex:word_match(lexer.FUNCTION_BUILTIN))
-local func = lex:tag(lexer.FUNCTION, lexer.word)
-local method = (B('.') + B('->')) * lex:tag(lexer.FUNCTION_METHOD, lexer.word)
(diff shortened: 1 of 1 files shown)


orbitalquark:
Thanks for submitting this. Since it looks like C3 inherits much of its syntax from C, instead of copy-pasting the C lexer and editing word lists, I think it's better to use `local lex = lexer.new(..., {inherit = lexer.load('c')})` and then `lex:set_word_list()` with C3's keywords, types, etc. The *lexers/glsl.lua* is an example. Note that `lex:set_word_list()` has a third parameter for appending to an existing word list that hopefully works for inherited lexers. Let me know if it does not.

orbitalquark:
Thanks for changing the lexer to inherit from the C lexer. It looks better.

I was studying the C3 specification, and it's pretty inconsistent (for example, keywords in https://c3-lang.org/implementation-details/grammar/ vs https://c3-lang.org/implementation-details/specification/).

The 'reflection syntax' you mentioned in 0.7.x appears to only be "methods" for types (e.g. `int.alignof`), and thus not true keywords. These feel more like C's methods (and would be tagged as such). However, it appears that 0.8.x will switch to `int::alignof` syntax, so the language appears to be undergoing significant changes.

(By the way, your `lexer:add_rule()` calls should be replaced by `lexer:modify_rule()` calls, otherwise the C lexer's rules will take precedence and you won't get the expected results)

All this being said, I'm not sure how I feel about committing this lexer given the state of flux in the C3 language. Perhaps it would be worth keeping this PR open and holding off to see how the language solidifies. Also, the self-described "basic and jank"-iness of this lexer doesn't instill much confidence, as it'll likely become a maintenance burden on me down the road.

nubbsterr:
Yea they literally updated the docs sometime last week (?) and it's genuinely annoying. I myself probably won't touch this for a bit until a 1.0 release (which is scheduled for 2028!!?)

I'm on board w/ keeping this open for *now* though it would also be fine to just close it *until another 2 years pass* once this language is (hopefully) stable.

PR #204, opened 2026-05-26
fix links in docs

Udi-Fogiel:
there is also index.html#contribute in manual.md, not sure where it should point to.


Udi-Fogiel:
Just realized the links are for the io pages. Sorry for the noise.

PR #208, opened 2026-06-03
Improve Perl heredoc lexing

1-1sam:
This PR changes the behavior of the Perl heredoc lexer to accept heredocs anywhere on a line, rather than exclusively at the end of a line/statement.

### DIFF
--- a/lexers/perl.lua
+++ b/lexers/perl.lua
@@ -56,7 +56,7 @@ local sq_str = lexer.range("'")
 local dq_str = lexer.range('"')
 local cmd_str = lexer.range('`')
 local heredoc = '<<' * P(function(input, index)
-	local s, e, indented, _, delimiter = input:find('(~?)(["\'`]?)([%a_][%w_]*)%2[\n\r\f;]+', index)
+	local s, e, indented, _, delimiter = input:find('(~?)(["\'`]?)([%a_][%w_]*)%2.-[\r\n]+', index)
 	if s == index and delimiter then
 		local end_heredoc = (#indented > 0 and '[\n\r\f]+ *' or '[\n\r\f]+')
 		e = select(2, input:find(end_heredoc .. delimiter, e))


lexers/perl.lua
```diff
@@ -56,7 +56,7 @@ local sq_str = lexer.range("'")
 local dq_str = lexer.range('"')
 local cmd_str = lexer.range('`')
 local heredoc = '<<' * P(function(input, index)
-	local s, e, indented, _, delimiter = input:find('(~?)(["\'`]?)([%a_][%w_]*)%2[\n\r\f;]+', index)
+	local s, e, indented, _, delimiter = input:find('(~?)(["\'`]?)([%a_][%w_]*)%2.-\n', index)
```
orbitalquark: How about `.-[\n\r]+`? That way we consume CRLF newlines.

Instead of the blanket `.-`, would it make sense to include a set of characters that would likely come next, and then `.-[\n\r]+`? Like a comma, space, etc. for example. I'm mainly thinking that an invalid character like an alphanumeric shouldn't result in matching a heredoc.

1-1sam:
I've made the changes you suggested. One issue with the new lexer is it won't be able to lex heredocs at the end of a line with no semicolon, operator, etc. I don't know if that would be considered an acceptable tradeoff. Perhaps we could have one `input:find` to match a heredoc followed by nothing but whitespace, and then if that does not match try the existing `input:find` pattern.

orbitalquark:
Hmm, I think I was wrong to suggest a trailing character because now it's required. If we made it optional, then an alphanumeric could still match. I think we should go back to `.-[\r\n]+`. Would that match a heredoc at the start of a line? (Sorry, I cannot test it out myself right now.)

1-1sam:
Reverted to the blanket `.-`. Heredocs at the start of a line seem to be matched fine.

orbitalquark:
Thank you! I took the liberty of improving some of the other heredoc matching in lexers that would benefit from this, like Bash and Ruby. https://github.com/orbitalquark/scintillua/commit/6182f7b70a85345fb12b34c1d5760ff3436230f3

PR #209, opened 2026-06-11
C99 functions/macros

1-1sam:
PR adds some missing functions and macros from C99.


orbitalquark:
Thank you! Committed in https://github.com/orbitalquark/scintillua/commit/309d7ea36bf11666d6a9e78589bf5fa8fb0998ec

PR #210, opened 2026-06-17
Add Odin lexer

LibreCoder951:
This PR adds a lexer for Odin.
I saw there already was one attempt years ago, but the PR is old and already closed anyway, so here's a new one.

Let me know if you would like to see any improvements or changes.

image

### DIFF
--- a/lexers/odin.lua
+++ b/lexers/odin.lua
@@ -12,6 +12,23 @@ local P, S = lpeg.P, lpeg.S
 local lex = lexer.new(...)
 
 -- Rules.
+local sq_str = P('L')^-1 * lexer.range("'")
+local dq_str = P('L')^-1 * lexer.range('"')
+local ml_str = P('L')^-1 * lexer.range('`')
+lex:add_rule('string', lex:tag(lexer.STRING, sq_str + dq_str + ml_str))
+
+local comment = lexer.to_eol('//', true)
+local block_comment = lexer.range('/*', '*/')
+lex:add_rule('comment', lex:tag(lexer.COMMENT, comment + block_comment))
+
+local directive = lex:tag(lexer.PREPROCESSOR, '#' * lexer.word)
+lex:add_rule('directive', directive)
+
+local attribute = lex:tag(lexer.PREPROCESSOR,
+	'@(' * (lexer.word + lexer.number + lexer.range('"') + lexer.range("'") + P('=') + P(' '))^0 * ')'
+)
+lex:add_rule('attribute', attribute)
+
 lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD)))
 lex:add_rule('type', lex:tag(lexer.TYPE, lex:word_match(lexer.TYPE)))
 lex:add_rule('constant', lex:tag(lexer.CONSTANT, lex:word_match(lexer.CONSTANT)))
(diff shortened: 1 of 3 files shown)


lexers/odin.lua
```diff
@@ -0,0 +1,110 @@
...
+
+local lex = lexer.new(...)
+
+-- Rules.
+local sq_str = P('L')^-1 * lexer.range("'")
+local dq_str = P('L')^-1 * lexer.range('"')
+local ml_str = P('L')^-1 * lexer.range('`')
+lex:add_rule('string', lex:tag(lexer.STRING, sq_str + dq_str + ml_str))
+
+local comment = lexer.to_eol('//', true)
```
orbitalquark, line 20: '#!' is also a valid line comment prefix.

I see that file tags like `#+vet !using-stmt !using-param` are a thing. I'd like to include them here if that's okay with you.

lexers/odin.lua
```diff
@@ -0,0 +1,110 @@
...
+local block_comment = lexer.range('/*', '*/')
+lex:add_rule('comment', lex:tag(lexer.COMMENT, comment + block_comment))
+
+local directive = lex:tag(lexer.PREPROCESSOR, '#' * lexer.word)
+lex:add_rule('directive', directive)
+
+local attribute = lex:tag(lexer.PREPROCESSOR,
```
orbitalquark, line 27: `@private` is valid, so we should allow `'@' * lexer.word` too.

Also, I think as written, this is too restricted. I propose `'@' * lexer.word^-1` so that anything inside `@(...)` is highlighted as odin code.

lexers/odin.lua
```diff
@@ -0,0 +1,110 @@
...
+	'@(' * (lexer.word + lexer.number + lexer.range('"') + lexer.range("'") + P('=') + P(' '))^0 * ')'
+)
+lex:add_rule('attribute', attribute)
+
+lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD)))
+lex:add_rule('type', lex:tag(lexer.TYPE, lex:word_match(lexer.TYPE)))
+lex:add_rule('constant', lex:tag(lexer.CONSTANT, lex:word_match(lexer.CONSTANT)))
+lex:add_rule('function', lex:tag(lexer.FUNCTION, lex:word_match(lexer.FUNCTION)))
```
orbitalquark, line 35: I think a `#P('(')` suffix would be useful here too.

Also, a `-lpeg.B('.')` prefix would avoid matching library functions as builtin functions (e.g. something like `vector.abs()`).

Finally, this should be `lexer.FUNCTION_BUILTIN`

lexers/odin.lua
```diff
@@ -0,0 +1,110 @@
...
+lex:modify_rule('constant', lex:get_rule('constant') + lex:word_match(lexer.VARIABLE_BUILTIN))
+lex:add_rule('number', lex:tag(lexer.NUMBER, lexer.number))
```
orbitalquark, line 37: A `S('ijk')^-1` suffix allows for imaginary numbers.

lexers/odin.lua
```diff
@@ -0,0 +1,110 @@
...
+lex:add_rule('directive', directive)
+
+local attribute = lex:tag(lexer.PREPROCESSOR,
+	'@(' * (lexer.word + lexer.number + lexer.range('"') + lexer.range("'") + P('=') + P(' '))^0 * ')'
+)
+lex:add_rule('attribute', attribute)
+
+lex:add_rule('keyword', lex:tag(lexer.KEYWORD, lex:word_match(lexer.KEYWORD)))
+lex:add_rule('type', lex:tag(lexer.TYPE, lex:word_match(lexer.TYPE)))
+lex:add_rule('constant', lex:tag(lexer.CONSTANT, lex:word_match(lexer.CONSTANT)))
```
orbitalquark, line 34: This should be `lexer.CONSTANT_BUILTIN`.

orbitalquark (commented):
Thanks for submitting this! I left some comments below that I'd like to get your OK/opinion on. I have made the necessary changes locally, so there's no need for you to make any changes yourself. I left the copyright attribution alone, so I want to know if you're okay having your name attached to my proposed changes. Otherwise let me know what you prefer.

LibreCoder951:
Yes, those look fine to me. Let me know if there's anything else.

orbitalquark:
Committed in https://github.com/orbitalquark/scintillua/commit/73b9f0a7fe70e307e71d8b3607ed1b2697ebfd14. I also did some rule rearrangement. Thank you!

PR #211, opened 2026-06-19
More C99 functions/constants

1-1sam:
This PR adds some more missing C99 functions/constants, mostly relating to the `long long`, `unsigned long long`, `float`, and `long double` types.


orbitalquark:
Thanks. Committed in https://github.com/orbitalquark/scintillua/commit/275d302dac22053a5aae27d9eab678008420c672

PR #212, opened 2026-06-26
C: Move enum to type keyword list

1-1sam:
This PR moves `enum` to the type keyword list as opposed to the normal keyword list. This is because I believe `enum` has more in common with compound type keywords like `struct` or `union` than other normal keywords. As an example, the `enum` type in the screenshot below would be green instead of yellow, which would be more consistent with the other compound types.

Screenshot_2026-06-26_09-47-29

What are your thoughts on this?  I understand this might be more of a case of personal taste.

### DIFF
--- a/lexers/c.lua
+++ b/lexers/c.lua
@@ -71,7 +71,7 @@ lex:add_fold_point(lexer.COMMENT, '/*', '*/')
 
 -- Word lists.
 lex:set_word_list(lexer.KEYWORD, {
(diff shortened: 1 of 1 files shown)


orbitalquark:
I think this is reasonable, considering the C++ lexer does this. Committed in https://github.com/orbitalquark/scintillua/commit/b9986ecad77b1ea73d75bf1e82e6e0fd3b4958b1

PR #213, opened 2026-07-27
Recognize wchar.h and wctype.h

1-1sam:
This PR adds [wchar.h](https://en.cppreference.com/c/header/wchar) and [wctype.h](https://en.cppreference.com/c/header/wctype) functions, types, and constants.


orbitalquark:
Thanks! Committed in https://github.com/orbitalquark/scintillua/commit/8a272df8507bc27bda1967953e200374908af3a9.

